PR #26, opened 2024-06-27
Remove runApi and getFormatedResponse

ChristopherPark3:
- Removes calls to runApi and getFormatedResponse functions. 
- Refactors api functions to interface with axios directly and allow for RQ to handle errors directly from axios as opposed to our formatted responses.

- Just had minor issues with the shape of the data the frontend was expecting. 
- Tested each RQ hook and fixed these errors.

- Less overhead when making an api call.
- Also, as we had it, the formatted response wouldn't throw an error so we weren't taking full advantage of the RQ hooks. 
- Now, we can just simply make the axios call in the RQ hook and RQ will handle error states for us.
  - Useful in useMutation 'onError' callbacks as well as error states that are available from useQuery and useMutation hooks.

- Simply refactors all functions that are making api calls to call axios directly as opposed to abstracting it in 'runApi'.
- Also make sure that exceptions are thrown in directly in RQ hooks so that we can take advantage of RQ features.

# Checklist

- [ x ] I have commented my code, particularly in hard-to-understand areas.
- [ x ] Already rebased against main branch.

# Screenshots


src/api/auth.ts
```diff
@@ -82,7 +100,7 @@ export const checkIsAccountAvailable = async (username: string) => {
     }
   } catch (error: any) {
     // Error in checking availabilty, retry
-    console.error('Error checking account availability:', error);
-    return false;
+    // console.error('Error checking account availability:', error)
```
Cutaiar, line 103: Remove comment

Cutaiar, line 103: Remove comment

src/api/gigs.ts
```diff
@@ -1,143 +1,180 @@
...
   const reqBody = { status };
-  const res = await runApi(
-    'PATCH',
-    `${API_URL}/milestones/${milestoneId}/status`,
-    reqBody,
-    {},
-    true
-  );
-  return res;
+  const apiReqHeader = validateToken({}, true);
```
Cutaiar, line 140: Suggest calling this **and all other occurrences** `reqHeader`. Is there any reason to include `api`?

```suggestion
  const reqHeader = validateToken({}, true);
```

src/api/gigs.ts
```diff
@@ -1,143 +1,180 @@
...
+
+  try {
+    const res = await axios.patch(
+      `${API_URL}/offers/status/${offerId}`,
+      reqBody,
+      apiReqHeaders
+    );
+    return res;
+  } catch (error: any) {
+    throw new Error('Failed to reject offer');
```
Cutaiar, line 166: I like that the thrown error includes "Failed to reject offer". Other thrown errors use just `error.response.data.message`. Have you considered using both?

```suggestion
    throw new Error(`Failed to reject offer: ${error.response.data.message}`);
```

I think using both helps us understand **what happened** and **why**. This is relevant for **all other instances** of thrown errors too!

src/app/gigs/gigs-comms.tsx
```diff
@@ -94,18 +96,26 @@ export const GigsComms = ({
...
   } = useGetContractForOffer({
     currentOfferId,
     isOpen,
     activeTab,
   });
 
+  useEffect(() => {
+    toast({
+      title: 'Unable to fetch contract information',
+      description: 'If the problem persists, please contract support',
```
Cutaiar, line 115:
```suggestion
      description: 'If the problem persists, please contact support',
```

When we ask people to contact support, there should be a link for them to do so. Maybe we should chat with Rami about this?

src/components/signin/signin-form.tsx
```diff
@@ -4,12 +4,12 @@ import { Button } from '../ui/button';
 import { Input } from '../ui/input';
 import { Label } from '../ui/label';
 import { FormContainer } from './form-container';
-import { getUserByAccountId } from '@/api/user';
+// import { getUserByAccountId } from '@/api/user';
```
Cutaiar, line 7: Remove rather than comment

src/components/signin/signin-form.tsx
```diff
@@ -81,13 +86,7 @@ const SignInForm = () => {
...
         if (activeAccount) {
           const response = await getUserByAccountId(activeAccount);
-          if (response.status !== 'success') {
-            toast({
-              title: 'No account found',
-              description:
-                'No account found for the connected wallet, please signup first',
-              variant: 'destructive',
-            });
+          if (response.data == 'failed') {
```
Cutaiar, line 89:
```suggestion
          if (response.data === 'failed') {
```

src/hooks/auth/useLoginWithEmail.ts
```diff
@@ -7,14 +7,7 @@ export const useLoginWithEmail = () => {
...
   return useMutation({
-    mutationFn: async (email: string) => {
-      const response = await loginWithEmail(email);
-      if (response.status === 'success') {
-        return response;
-      } else {
-        throw new Error(response.message || 'Failed to login with email');
-      }
-    },
+    mutationFn: async (email: string) => await loginWithEmail(email), // Make sure to update loginWithEmail in useAuth
```
Cutaiar, line 10: is `loginWithEmail` updated?

src/hooks/gigs/useFetchOffers.ts
```diff
@@ -1,10 +1,49 @@
-import { useAuth } from '../useAuth';
 import { getClientOffers, getFreelancerOffers } from '@/api/gigs';
 import { useToast } from '@/components/ui/use-toast';
 import { GigsContentType, User } from '@/types';
-import { FormattedResponse } from '@/types';
 import { useQuery } from '@tanstack/react-query';
 
+// export const useFetchOffers = (
```
Cutaiar, line 6: Consider removing commented blocks

src/hooks/gigs/useFetchOffers.ts
```diff
@@ -29,14 +68,17 @@ export const useFetchOffers = (
...
-        if (response && response.status === 'success') {
+
+        if (response.data && response.statusText === 'OK') {
           return response.data;
         } else {
           toast({
-            title: 'Unable to load offers',
-            description: 'Contact support if issue persists',
+            title: 'Failed to fetch offers',
+            description: 'Try again. If this issue persists, contact support.',
```
Cutaiar, line 77: Same comment as above for a support link.

src/utils/icons.tsx
```diff
@@ -31,7 +31,8 @@ export const parseWidgetTypeFromUrl = (url: string): WidgetType => {
     const urlObj = new URL(url);
     const hostname = urlObj.hostname;
     const pathname = urlObj.pathname;
-    const domainParts = hostname.split('.');
+    const domainParts = hostname.split('.')
+    console.log('domainParts', domainParts)
```
Cutaiar, line 35: remove console log

Cutaiar (approved):
Looks good overall. I left a few comments which apply to most of the changes. The main thing I am concerned about here is when you check for a successful response with:

```ts
if (response && response.data) {
    // Do something with response
}
``` 

and various other ways. Like `response.data.statusText` etc. As long as you have tested these, and I think this is good to go through!

Nice work :).

src/api/gigs.ts
ChristopherPark3: No specific reason other than following the prior naming conventions, I'll go ahead and change the occurrances to `reqHeader`!

src/api/gigs.ts
ChristopherPark3: Yea I think that makes a lot of sense! I think I got lazy and trusted the error messages that axios throws a little too much. That's my bad and will go back and make it more detailed!

PR #31, opened 2024-07-17
feat(auth): add FastAuth for signin and signup

wonderboy25:
Near Fast auth sign in/up


Cutaiar:
Looks like this needs to be merged with current `main`. @wonderboy25 would you like me to make a few commits resolving this?

wonderboy25:
yes, let me know about what parts need plz

wonderboy25:
Done!

src/components/Header/header.tsx
```diff
@@ -18,7 +17,7 @@ export const Header = ({ isPublic = false }: HeaderProps) => {
       <div className='flex w-full justify-between container mx-auto py-4'>
         <div className='flex gap-6'>
           <Link href='/'>
-            <Image src='/svg/logo.svg' alt='logo' width={44} height={44} />
+            <img src='/svg/logo.svg' alt='logo' width={44} height={44} />
```
Cutaiar, line 20: Is there a reason for changing this to `` tag? Next `<Image>` gets us [nice optimizations ](https://nextjs.org/docs/app/building-your-application/optimizing/images) for free.

wonderboy25: The reason is SEO and Performance.
Search engines can index and understand  tags more effectively, potentially improving SEO for your website. Additionally, browsers optimize loading and rendering of  tags, especially when using attributes like srcset and sizes for responsive images.

package-lock.json
Cutaiar, line 1: This file should not be committed. We are installing deps with yarn.

Cutaiar:
Added a few commits which include changes that this PR was reverting. They are irrelevant to fastauth, but maybe one more quick test is worth doing.

LICENSE
Cutaiar, line 1: This license does not apply to the entire codebase. It should be edited or removed.

package-lock.json
Cutaiar, line 1: This file should not be committed. We are installed deps with yarn.

package.json
```diff
@@ -3,7 +3,7 @@
   "version": "0.1.0",
   "private": true,
   "scripts": {
-    "start:dev": "PORT=6100 next dev",
```
Cutaiar, line 6: Is there a reason for changing the port? If not, let's keep it at 6100.

src/components/signup/all-set.tsx
```diff
@@ -31,7 +31,7 @@ const AllSet = () => {
         </div>
         <Button
           className='w-full text-white bg-[#573DF5] border border-[#7F56D9] shadow-sm shadow-[#1018280D] font-semibold text-base'
-          onClick={() => router.push('/')}
+          onClick={() => router.push('/daniel')}
```
Cutaiar, line 34: This seems like a WIP change. Should this be:

```suggestion
          onClick={() => router.push('/')}
```

src/hooks/gigs/useGetContractForOffer.ts
```diff
@@ -28,7 +29,7 @@ export const useGetContractForOffer = (data: useGetContractForOfferParams) => {
         throw new Error('Unable to fetch contract information');
       }
     },
-    // Query function will only run when the isOpen and activeTab are true and the activeTab is the Contract tab
+    // Query function will only run whenm the isOpen and activeTab are true and the activeTab is the Contract tab
```
Cutaiar, line 32:
```suggestion
    // Query function will only run when the isOpen and activeTab are true and the activeTab is the Contract tab
```

wonderboy25:
Pushed a resolved commit

Cutaiar:
Thanks for your work on this. I'll leave the remaining review and approval to @humzakhan

Cutaiar (commented):
What are the styled components introduced for? 

```
lib/Badge/
    Button/
    Spinner/
    Text/
    Toast/
    Tooltip/
```

Doesn't seem like they are used anywhere. Are they an artifact of branching off an early commit to main? If they are not needed, please remove them. If they are needed, it would be better to use [shadcn](https://ui.shadcn.com/) components with tailwind styles as that is what the rest of the repo uses. Let me know if including these is intentional and if so, point me to where they are used and I would be happy to convert them to shadcn components.

src/api/index.ts
```diff
@@ -0,0 +1,222 @@
+import { withTimeout } from '../utils';
+import { network, networkId } from '../utils/config';
+import { CLAIM, getUserCredentialsFrpSignature } from '../utils/mpc-service';
+import { LimitedAccessKey, NewAccountResponse } from './types';
+import { captureException } from '@sentry/react';
+import { KeyPair } from 'near-api-js';
+
```
Cutaiar, line 7: would prefer if this file was called `fastAuthUtils.ts` or something a little more descriptive.

src/api/types.ts
```diff
@@ -0,0 +1,22 @@
+export type LimitedAccessKey = {
```
Cutaiar, line 1: similarly, this file could be `fastAuthTypes.ts` or something like that

src/components/ui/input.tsx
```diff
@@ -2,23 +2,34 @@ import { cn } from '@/lib/utils';
...
-        type={type}
-        className={cn(
-          'flex h-10 w-full rounded-md border border-input bg-background px-3 py-2 text-sm ring-offset-background file:border-0 file:bg-transparent file:text-sm file:font-medium placeholder:text-muted-foreground disabled:cursor-not-allowed disabled:opacity-50',
-          className
+      <div className='relative flex items-center'>
+        <input
+          type={type}
+          className={cn(
+            'flex h-10 w-full rounded-md border border-input bg-background px-3 py-2 text-sm ring-offset-background file:border-0 file:bg-transparent file:text-sm file:font-medium placeholder:text-muted-foreground disabled:cursor-not-allowed disabled:opacity-50',
+            suffix ? 'pr-10' : '', // Add padding to the right if there is a suffix
```
Cutaiar, line 17: You can do `suffix && 'pr-10'` since cn accepts falsy values

src/hooks/gigs/useGetContractForOffer.ts
```diff
@@ -15,12 +16,23 @@ type useGetContractForOfferParams = {
...
 export const useGetContractForOffer = (data: useGetContractForOfferParams) => {
   const { currentOfferId, isOpen, activeTab } = data;
+  const { toast } = useToast();
 
   const query = useQuery({
     queryKey: ['contractForOffer', currentOfferId],
     queryFn: async () => {
-      const res = await getContractForOffer(currentOfferId);
-      return res.data;
+      console.log('getContractOffer');
```
Cutaiar, line 24: please remove console log if not using

src/hooks/useAuthState.ts
```diff
@@ -0,0 +1,79 @@
...
+    }
+  } catch (e) {
+    captureException(e);
+    return false;
+  }
+
+  return false;
+};
+
+export const useAuthState = (): {authenticated: AuthState} => {
```
Cutaiar, line 65: looks like this should be `useFastAuthState`, the fn above should be `getFastAuthState` and the file should be `useFastAuthState.ts`

src/lib/config.ts
```diff
@@ -83,6 +91,10 @@ export const config: AppConfig = appConfigSchema.parse({
   spotifyClientSecret: process.env.NEXT_PUBLIC_SPOTIFY_CLIENT_SECRET,
   youtubeClientId: process.env.NEXT_PUBLIC_YOUTUBE_CLIENT_ID,
   youtubeClientSecret: process.env.NEXT_PUBLIC_YOUTUBE_CLIENT_SECRET,
+  loginSuccessUrl: process.env.NEXT_PUBLIC_LOGIN_SUCCESS_URL,
```
Cutaiar, line 94: I know this doesn't follow the current convention, but soon the rest will be fixed. Please use the same in .env as you use here. So this would be `NEXT_PUBLIC_LOGIN_SUCCESS_URL` and similar for the rest.

src/lib/controller.ts
```diff
@@ -0,0 +1,559 @@
+import { Account, Connection } from '@near-js/accounts';
```
Cutaiar, line 1: Seems like this file could be called `fastAuthController.ts`. I think that it either makes sense to do these file renames I have mentioned. Or keep their names if they are all in some kind of `fastAuth` folder that has just what is related to fastauth.

src/lib/networkParams.ts
```diff
@@ -0,0 +1,21 @@
+export default {
```
Cutaiar, line 1: this could be combined with the networks object in `./config.ts`

src/utils/analytics.ts
Cutaiar, line 1: All of the files added to `util/` here should be grouped into a `fastauth` folder and exported from an `index.ts` in that folder. Want to maintain separation of concerns and have fasuth be more like a module that our code is using. Ideally, it could even be a separate npm package one day...

src/utils/index.ts
```diff
@@ -1,24 +1,145 @@
 'use client';
 
+import { FormattedResponse } from '../types';
 import { config } from '@/lib/config';
+import axios from 'axios';
 
 export const API_URL = config.devApiUrl;
 
+export const getFormatedResponse = (res: any): FormattedResponse => {
```
Cutaiar, line 9: these were removed in a previous PR. Not sure why merging main in did not get rid of them. Please remove these two fns.

src/lib/config.ts
wonderboy25: I can't know what you mean, would you confirm again?

Cutaiar:
```suggestion
  NEXT_PUBLIC_LOGIN_SUCCESS_URL: process.env.NEXT_PUBLIC_LOGIN_SUCCESS_URL,
```
And the same for the other 3 below.

The change will need to be made in the zod and type definition above too.

src/lib/networkParams.ts
Cutaiar: It's okay if not, seems like some of our types come from the same place...

src/lib/config.ts
wonderboy25: I dont agree, because it is original format, 
If you want to change it, you can do it later

wonderboy25: All other config paras are same format too

src/utils/fastAuth/analytics.ts
Cutaiar, line 1: These analytics are not used anywhere. If you don't need them, please remove this file and `rudder-sdk-js`

Cutaiar (approved):

PR #34, opened 2024-07-23
Sendbird chat

ChristopherPark3:
** SENDBIRD COMMITS START ON JULY 16
  - Having an issue with old commits making it into my new branches.
 
- Implements Sendbird chat feature:
    - Chat between client and freelancer 
    - Text messages
    - File messages
    - Typing indicator
    - Slight tweaks to ui to account for new features

- Features to implement in other smaller PRs:
    - Session tokens (backend functions are written already)
    - Read only chats after gigs are completed (with 24 hour grace window)
    - Knock syncing

- No major problems at this point. Just had a lot of trouble with managing and manipulating files. Took the longest.
- Having trouble with a weird ui bug of the chat overflowing outside of its container
- The ui for the chat, specifically file messages is just a temporary ui until we get a mock up from the design team

- Allows clients and freelancers to chat and communicate about important contract data

- Keep users in Sorbet as much as possible.

- Implemented the Sendbird SDK to facilitate chat communication between users.
- Established web socket connections when a gig is clicked and the chat component is loaded. 

# Checklist

- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] Already rebased against main branch.

# Screenshots

https://www.loom.com/share/bc7b133409ca4c4b8b813ae4083218e0?sid=a45a9c72-b67f-47cd-b06e-2e22c39783d5


src/components/chat/typing-indicator.tsx
Cutaiar, line 1: I think this should go in `src/app/gigs/chat`. with the rest of the UI. You'll see in my widget PR's, I am trying to colocate code for components. As far as I see, this is only used in the components in `app/gigs/chat` so it should stay there.

IMO, we can use `src/components` for components that are _truly shared_.

src/api/chat.ts
```diff
@@ -0,0 +1,27 @@
+// 'use server';
```
Cutaiar, line 1: Did you mean to leave this?

src/components/chat/typing-indicator.tsx
ChristopherPark3: Gotcha sounds good I'll make that change!

src/api/chat.ts
ChristopherPark3: No, that was by accident I will remove it now!

src/utils/sendbird.ts
Cutaiar, line 1: Similar to my other comment, this is not used or useful outside of the sendbird components. Therefore, it makes sense to colocate it with those components. Most of the stuff in this util directory will have the same fate. The only thing that should go here is utils that are used by many components.

ChristopherPark3: Gotcha that makes sense. I'll move this file for now, but will keep a note to move the rest, probably after beta launch

src/app/gigs/chat/chat-bottombar.tsx
```diff
@@ -1,69 +1,104 @@
...
-      const newMessage: Message = {
-        id: message.length + 1,
-        name: loggedInUserData.name,
-        avatar: loggedInUserData.avatar,
-        message: message.trim(),
-      };
-      sendMessage(newMessage);
-      setMessage('');
+    if (message.length > 0) {
+      if (message.trim()) {
```
Cutaiar, line 83: What is the purpose of checking the return value of `message.trim()`? And then sending the original `message`?

src/app/gigs/chat/chat-file-preview.tsx
Cutaiar, line 1: You have probably already seen, but just want to drop a note that [SendBirds UIKit](https://sendbird.com/docs/chat/uikit/v3/react/features/file-sharing) speed up an implementation of file previews a lot when it is designed.

src/app/gigs/chat/chat-layout-minimal.tsx
```diff
@@ -1,25 +1,151 @@
...
+  const channelHandlers = new GroupChannelHandler({
+    onTypingStatusUpdated: (groupChannel: GroupChannel) => {
+      const typingUsers = groupChannel.getTypingUsers();
+      updateState({ ...stateRef.current, typingMembers: typingUsers });
+    },
+  });
+
+  useEffect(() => {
+    async function initializeChat() {
+      await initializeConnection(user?.id);
```
Cutaiar, line 95: Do you want to initialize the connection and continue below if user is undefined?

Additionally, you might want to include user in the dep array if this effect.

We should be getting warnings about this via `eslint-plugin-react-hooks` but it looks like that might not be working rn.

src/app/gigs/gigs-comms.tsx
```diff
@@ -109,14 +109,14 @@ export const GigsComms = ({
   });
 
   useEffect(() => {
-    toast({
-      title: 'Unable to fetch contract information',
-      description: 'If the problem persists, please contract support',
-    });
+    if (isGetContractError === true) {
```
Cutaiar, line 112:
```suggestion
    if (isGetContractError) {
```

Right? since this is a boolean?

src/constant/constant.ts
Cutaiar, line 1: This file also is also going to leave. What do you think about storing sendbird config in `src/lib/config.ts` with the rest of the config information?

src/utils/sendbird.ts
```diff
@@ -0,0 +1,146 @@
...
+
+  return parseFloat((bytes / Math.pow(k, i)).toFixed(dm)) + ' ' + sizes[i];
+};
+
+const convertTimeToMinutes = (timeStr: string) => {
+  const [hours, minutes] = timeStr.split(':').map(Number);
+  return hours * 60 + minutes;
+};
+
+const getTimeDifferenceInMinutes = (time1: string, time2: string) => {
```
Cutaiar, line 130: This will depend on design, but a good package to be aware of is [`react-time-ago`](https://www.npmjs.com/package/react-time-ago). Makes it super easy to do local "5 min ago"

Cutaiar (approved):

src/app/gigs/chat/chat-bottombar.tsx
ChristopherPark3: That was my mistake, missed that and fixing now!

src/constant/constant.ts
ChristopherPark3: I like that, would we then just store all of our constants in one object in the config file?

Cutaiar: imo, there could be a case to separate _some_ constants from the app config. But no these. `LOCAL_KEY` is used in one place to set something in local storage. `ROLE_KEY` is not used at all. And `CONTACT_ID` is duplicated in the app config. Hence why this file should be removed.

ChristopherPark3: Gotcha sounds good, I just added it there for consistency. I'll go ahead and move the sendbird stuff to the config and set that up.

PR #37, opened 2024-07-26
u/cp/sendbird knock

ChristopherPark3:
** STILL NEEDS TO BE MERGED WITH THE `u/cp/claim-sorbet` repo. Waiting for that to be reviewed.
** COMMITS START JULY 25TH

- Implements users receiving message notifications when they are away from chat. 
- Note, double messages will appear in dev because of React strict mode. I tested in prod and only one notification appeared.
    - The cause is because I am running an effect to check whenever the `Dialog` component opens and closes. When it closes, I then disconnect from the Sendbird server.
- Also implements ui changes from Chris Astles in regards to the display of file messages.

- The only problem that I see is the Knock messages not appearing how I expect. I think it has something to do with the Knock workflow and editing the message in the Knock dashboard should fix it.

- Allows users to receive message notifications when they are away from their keyboard

- Incorporate messaging into Knock so that users can always know when others respond or reach out.

- Everytime a user closes out of a chat, we are considering them `offline` and do so by disconnecting from the Sendbird servers. We do this so that the 'connectionStatus' for each Sendbird User object is properly updated.
- Then, whenever a new message is added, we run a check to see if the other user is `offline` and if so, we make a call to our backend to trigger the Knock workflow to send a notification.

# Checklist

- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] Already rebased against main branch.

# Screenshots

UI changes for file messages:
Screenshot 2024-07-25 at 7 47 29 PM

What the notification text looks like in app:
Screenshot 2024-07-25 at 7 48 05 PM

What it looks like in the Knock dashboard: 
Screenshot 2024-07-25 at 7 48 59 PM


src/app/gigs/gigs-comms.tsx
```diff
@@ -108,6 +109,17 @@ export const GigsComms = ({
...
+  // This effect is mainly to disconnect from Sendbird so that when a new message is added, the connectionStatus property is
+  // properly being updated when a user closes out of the chat
+  useEffect(() => {
+    async function disconnectSendbird() {
+      if (!isOpen) {
+        await removeConnection();
+      }
+    }
+    disconnectSendbird();
+  }, [isOpen]);
```
humzakhan, line 121: You can move the check out as following:

```
useEffect(() => {
  async function disconnectSendbird() {
    await removeConnection();
  }

  if (!isOpen) {
    disconnectSendbird();
  }
}, [isOpen]);
```

src/app/gigs/chat/chat-layout-minimal.tsx
```diff
@@ -46,7 +53,25 @@ export function ChatLayoutMinimal({
   stateRef.current = state;
 
   const messageHandlers: MessageCollectionEventHandler = {
-    onMessagesAdded: (context: any, channel: any, messages: any) => {
+    onMessagesAdded: async (context: any, channel: any, messages: any) => {
+      // a check to see if the recipient is offline or online
+      // connectionStatus is determined if there is an active connection to Sendbird
+      const recipient = channel.members.find(
+        (member: any) => member.userId !== user?.id
```
humzakhan, line 60: The value of recipient will be the first object in `channel.members` with a different userId and we have no way to be sure it's the actual recipient. 

We need to get the recipient by the actual userId.

src/app/gigs/gigs-comms.tsx
ChristopherPark3: Ahh that's a lot more elegant, will do!

src/app/gigs/chat/chat-layout-minimal.tsx
ChristopherPark3: Yea that makes a lot of sense, I did it like this because there will only ever be 2 users in a channel, but I will switch to the filter method to be safe!

src/app/gigs/chat/chat-file-display.tsx
```diff
@@ -11,21 +12,35 @@ import {
...
-  handleOpenFile: () => void;
   color: string;
 }
 
-const FileDisplay = ({ color, file }: FileDisplayProps) => {
+const icons = {
+  pdf: <File className='w-5 h-5 text-white' />,
+  jpeg: <ImageIcon className='w-5 h-5 text-white' />,
+  png: <ImageIcon className='w-5 h-5 text-white' />,
+};
```
Cutaiar, line 41: You can add 
```ts
type SupportedFileIcon = keyof typeof icons;
```
and use that below on line 111:
```ts
  file.type.split('/').pop()?.toLowerCase() as SupportedFileIcon
```
Follow up question: how do you know / validate that the file type is in fact one of those? Have you tested if it is not? Perhaps providing a generic "file" icon fallback?

src/app/gigs/chat/chat-file-display.tsx
```diff
@@ -75,27 +90,50 @@ const FileDisplay = ({ color, file }: FileDisplayProps) => {
...
-            {file.type.split('/').pop()}
-          </FileDisplay.Container>
+      <div className='flex flex-1 flex-row gap-2 '>
+        {!loading ? (
+          link ? (
+            <FileDisplay.Container className={color}>
+              <a href={link}>
+                <Image
+                  className='object-cover h-auto w-auto rounded-sm'
+                  src={link}
```
Cutaiar, line 100: Does this NextJS `<Image>` work> Usually it throws an error if you give it a src that's not from our domain.

src/app/gigs/chat/chat-layout-minimal.tsx
Cutaiar, line 1: line 149: channelID is always defined now, no need for if

src/app/gigs/chat/chat-layout-minimal.tsx
```diff
@@ -46,7 +57,30 @@ export function ChatLayoutMinimal({
   stateRef.current = state;
 
   const messageHandlers: MessageCollectionEventHandler = {
-    onMessagesAdded: (context: any, channel: any, messages: any) => {
+    // Channel will always be defined whenever this event is triggered bc it is triggered thru sendbird and
+    // only runs when a channel is successfully found between two users
+    onMessagesAdded: async (context: any, channel: any, messages: any) => {
```
Cutaiar, line 62:
```suggestion
    onMessagesAdded: async (context, channel, messages) => {
```
The type on the line above types these params. why override as `any`?

src/app/gigs/chat/chat-layout-minimal.tsx
```diff
@@ -46,7 +57,30 @@ export function ChatLayoutMinimal({
...
+      const recipientId = user?.id === clientId ? freelanceId : clientId;
+      // a check to see if the recipient is offline or online
+      // connectionStatus is determined if there is an active connection to Sendbird
+      const sender = channel.members.find(
+        (member: any) => member.userId === senderId
+      );
+      const recipient = channel.members.find(
+        (member: any) => member.userId !== recipientId
+      );
+      if (recipient.connectionStatus === 'offline') {
```
Cutaiar, line 74: If you remove the `any` types above, TS will show you that `recipient` is possibly undefined. This should be handled.

src/app/gigs/chat/chat-sidebar.tsx
```diff
@@ -13,11 +12,14 @@ import { cn } from '@/lib/utils';
 import { MoreHorizontal, SquarePen } from 'lucide-react';
 import Link from 'next/link';
 
+/**
+  Make sure to define the type of the messages prop
```
Cutaiar, line 16:
```suggestion
  TODO: Make sure to define the type of the messages prop
```
right?

Cutaiar (approved):
Approved with suggestions ✨

src/app/[username]/layout.tsx
```diff
@@ -1,5 +1,6 @@
 import { ReactNode } from 'react';
 
+//@ts-ignore
```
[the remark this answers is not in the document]
Comment (ChristopherPark3) [reply src/app/[username]/layout.tsx:3]: I was getting a a weird ts error. This was the first time I got it and when I compared this file to main to see if I accidentally changed anything, nothing was different. It was weird so I just ignored it for now.

src/app/gigs/chat/chat-file-display.tsx
ChristopherPark3: Its the `type` property from the `File` object I believe.  This is the data straight from the Sendbird storage and it was always consistent from when I was implementing the feature.  I'll work on the fallback icon now!

src/app/gigs/chat/chat-file-display.tsx
ChristopherPark3: The blob that's returned from `fetchFile` actually returns a blob that contains our domain name. So during testing, the return value was always `blob:http//localhost:6100/uniqueSendbirdFileId` and I didn't find anything else happening.

The only instance that I had to be aware of was when `fetchFile` was called on the server (Server Action) because it was returning `blob:node[something]:http://localhost...`

src/app/gigs/chat/chat-layout-minimal.tsx
ChristopherPark3: Nice catch thanks!

src/app/gigs/chat/chat-layout-minimal.tsx
ChristopherPark3: Ahh I was getting a linting error before, but now I just removed it and now the types are properly assigned! 👍

src/app/gigs/chat/chat-layout-minimal.tsx
ChristopherPark3: done! 🙌

src/app/[username]/layout.tsx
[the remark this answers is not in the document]
Comment (ChristopherPark3) [reply src/app/[username]/layout.tsx:3]: Screenshot 2024-07-26 at 4 26 56 PM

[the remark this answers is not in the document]
Comment (ChristopherPark3) [reply src/app/[username]/layout.tsx:3]: Screenshot 2024-07-26 at 4 27 21 PM

PR #39, opened 2024-07-28
feat(widgets): Add LinkedIn Profile Widget

Cutaiar:

Add basic support for LinkedIn Profile Widgets. LinkedIn is difficult to scrape, so for Testnet launch, just display their handle. Of course, clicking the widget links to their profile.

Sister: [sorbet-api](https://github.com/MySorbet/sorbet-api/pull/27)
# Screenshots
CleanShot 2024-07-27 at 23 23 53@2x


src/components/profile/widgets/widget-linkedin-profile.tsx
```diff
@@ -0,0 +1,125 @@
...
+            <Title>{name}</Title>
+          </WidgetHeader>
+          <BannerImage src={bannerImage} />
+        </WidgetLayout>
+      );
+    default:
+      throw new Error('Unsupported widget size');
+  }
+};
+
```
Cutaiar, line 71: All three local components below are copied from the newest Widget implementation. They should be shared between widgets, no need to review them here.

src/components/profile/widgets/widget.tsx
```diff
@@ -208,6 +210,30 @@ export const Widget: React.FC<WidgetProps> = ({
         );
         break;
 
+      case WidgetType.LinkedInProfile:
+        // For now, just say the name is their handle
```
Cutaiar, line 214: This is just a quick way to grab their handle. Right now, we don't even use widget content from the server

src/components/profile/widgets/widget.tsx
```diff
@@ -208,6 +210,30 @@ export const Widget: React.FC<WidgetProps> = ({
...
+        // TODO: Remove this once real linked in data is available
```
ChristopherPark3, line 215: So this is basically just the front end implementation? Waiting for backend?

ChristopherPark3 (approved):
Looks great! Simple and clean 🧼🫧

src/components/profile/widgets/widget.tsx
Cutaiar: Backend is implemented but does not make a request to linkedin. This is because they are very good at detecting scraping and most requests fail. Need more time to implement a scrape but in the mean time, users can have a linkedin widget with their handle rather than their name.

PR #43, opened 2024-08-06
Make chat available when a gig is created

ChristopherPark3:
- This PR allows for users to chat <strong>the moment a gig is created</strong>, as opposed to how we had it before where users had to wait <strong>until a contract was approved</strong>

# Implementation
- Created a `useSendbird` hook to store the initialized socket in state. 
- Updated `useChat` to connect based on data from `Offer` table instead of `Contract` table
- Creates `useGetOffers` and `getOffers` to fetch data for a specific offer (replaces the need for `contractData` in the chat component.)
    -> There can be an offer without a contract, but there can never be a contract and no offer. (the reason why we moved to chat exclusively interfacing with `Offer` table)

# Checklist:
[ x ] Make sure we are always making chat available to users (whether it is an offer or a contract has been approved)
[ x ] Small housekeeping: removing unused code in components that I directly worked with

# Screenshots

Provide screenshots or videos of the changes made if any (optional)


src/api/gigs.ts
```diff
@@ -196,3 +196,16 @@ export const createOffer = async (body: CreateOfferType) => {
     throw new Error(`Failed to create offer: ${error.response.data.message}`);
   }
 };
+
+export const getOfferById = async (offerId: string) => {
```
Cutaiar, line 200: Check out this quick video to understand how we can avoid the current antipatterns we have:

https://www.youtube.com/watch?v=NGSck4aHfeQ

src/hooks/chat/useChat.ts
```diff
@@ -25,19 +20,19 @@ import {
 import { MessageListParams } from '@sendbird/chat/message';
 import { useEffect, useRef, useState } from 'react';
 
-interface useInitializeChatProps {
+interface useChatProps {
   user: User | null;
   logout: () => void;
-  contractData: ContractType;
   isOpen: boolean;
+  offerData: PrismaOfferType | undefined;
```
Cutaiar, line 27: Consdider calling this `offer`. In general, the postfix `data` almost never helps provide more imformation about what this variable is or represents.

src/hooks/chat/useSendbird.ts
```diff
@@ -0,0 +1,81 @@
...
+} from '@sendbird/chat/groupChannel';
+import { useQuery } from '@tanstack/react-query';
+
+export const useSendbird = () => {
+  const { data: sb } = useQuery({
+    queryKey: ['sendbird'],
+    queryFn: async () => {
+      try {
+        const params: SendbirdChatParams<[GroupChannelModule]> = {
+          appId: config.sendbirdAppId as string,
```
Cutaiar, line 20: If you are doing this to assert it is not undefined, you can do `config.sendbirdAppId!`. We also want to use this very sparingly, but I understand why you need it. Ideally, `config.sendbirdAppId` is never undefined, and if it was, the app should fail much earlier. This will implemented soon and we can remove the assertion here.

src/hooks/chat/useChat.ts
```diff
@@ -47,7 +42,7 @@ export const useChat = ({
     typingMembers: [],
     messageCollection: null,
   });
-  console.log('useChat -> state', state);
+  const Sendbird = useSendbird();
```
Cutaiar, line 45: I would check something like `if (!Sendbird) { throw or fail }`. Rather than having each call be `Sendbird?`. If this init fails, there is no chat. Letting a null send bird object into the code below seems bug prone.

src/app/gigs/gigs-dialog.tsx
```diff
@@ -643,7 +628,10 @@ export const GigsDialog = ({
     <>
       <Dialog open={isOpen} onOpenChange={onOpenChange}>
         <DialogOverlay className='bg-[#F3F3F4]/90' />
-        <DialogContent className='flex flex-col md:h-[75vh] max-w-[900px] rounded-2xl'>
+        <DialogContent
+          className='flex flex-col md:h-[75vh] max-w-[900px] rounded-2xl'
+          aria-describedby={undefined}
```
Cutaiar, line 633: Reason this needs `aria-describedby` is becuase screen readers need to know what this dialog is called. I have a feeling the issue is that `DialogTitle` is not rendered as child of `Dialog`. There is a div in between. Consider moving the tw classes from the div to DialogTitle itself, moing the tab selector inside, and removing the div.

src/app/gigs/chat/chat.tsx
```diff
@@ -59,16 +39,15 @@ export function Chat({ showTopbar = true, contractData, isOpen }: ChatProps) {
...
         typingMembers={state.typingMembers}
-        contractStatus={contractData.status}
         supportedIcons={icons}
         chatLoading={chatLoading}
       />
       <div className='mt-4'>
         <ChatBottombar
           sendMessage={sendMessage}
-          isMobile={isMobile}
+          isMobile={false}
```
Cutaiar, line 48: Rather than pass isMobile false, I would have chat bottom bar decide if it is mobile using a hook, or better yet a container query. Need not be done for this PR.

src/app/gigs/chat/chat.tsx
```diff
@@ -17,40 +16,21 @@ type SupportedFileIcon = keyof typeof icons;
 
 interface ChatProps {
   showTopbar?: boolean;
-  contractData: ContractType;
   isOpen: boolean;
+  currentOfferId: string;
```
Cutaiar, line 20: consider calling this offerId?

src/app/gigs/chat/chat-bottombar.tsx
```diff
@@ -128,12 +128,12 @@ export default function ChatBottombar({
   };
 
   useEffect(() => {
-    if (
-      contractStatus ===
-      ('NotStarted' || 'InProgress' || 'InReview' || 'Completed' || 'Rejected')
-    )
+    //TODO: Figure out when chat becomes read only and sync the input with that disabled state
+    if (contractStatus === 'Completed') {
```
Cutaiar, line 132: would love to see `contractStatus` typed correctly via string union, but I see that is a much deepter issue. Not important for this PR.

Cutaiar (approved):

src/api/gigs.ts
ChristopherPark3: Thanks for linking that, I see now a lot of pitfalls that are currently in place... I'll clean up the calls used in this PR

src/app/gigs/chat/chat-bottombar.tsx
ChristopherPark3: Definitely will do!

PR #44, opened 2024-08-07
Clean up environment variables and tooling

Cutaiar:
# Clean up environment variables and tooling

- Remove unused variables from `env.example` and `config.ts`
- Remove makefile, moving the commands to `package.json`
- Dockerfiles specifically use yarn
- Use correct package name in `package.json`
- Remove `env-cmd` as next picks up everything from `.env.production` without this command
- Remove dotenv because it is similarly redundant
- Remove husky. We will not use precommit hooks.
- Clean up and centralize app config values
- Remove unused components
- Fixes a bug related to the paths of images not having the right case

Note: docker dev build did not work before this. It still does not.


ChristopherPark3 (approved):
Looks great!

PR #45, opened 2024-08-08
Call signup flow after fastauth user is created

Cutaiar:
# Description

- Refactor signup to not require a first and last name
- Call signup flow after fastauth user is created


ChristopherPark3 (approved):
Looks great!

PR #46, opened 2024-08-12
Widgets use string unions

Cutaiar:
# Description

- Convert WidgetType to string union
- Convert WidgetSize to string union


ChristopherPark3 (approved):
Looks great

PR #47, opened 2024-08-13
Fix Eslint

Cutaiar:
# Description
**Make sure you have the recommended VSCode plugins installed to take advantage of formatting and linting!**

- Updates `eslintrc.js` to be valid, thus allowing eslint to run during `next build`
- Since this whole app was built without eslint, configures some rules to warn, keeping the build working
- Fixes some obvious errors that eslint points out now that it is running. No functional changes.
- Refactors `validateToken` to `withAuthHeader` , adds a comment describing what it actually does, and removes useles parameter (we should use an [interceptor](https://blog.theashishmaurya.me/handling-jwt-access-and-refresh-token-using-axios-in-react-app#heading-setting-up-axios-interceptor-for-the-jwt-tokens) for this but for now, this is ok)

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -15,20 +15,21 @@ module.exports = {
   rules: {
     'no-unused-vars': 'warn',
     'no-console': 'warn',
-    '@typescript-eslint/explicit-module-boundary-types': 'off',
-    'react/no-unescaped-entities': 'off',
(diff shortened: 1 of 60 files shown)


ChristopherPark3 (approved):
We have lint! 🚀

PR #53, opened 2024-08-15
Updated header available everywhere

Cutaiar:
# Updated header available everywhere

Main change:
- Complete [SRBT-150: Hamburger -> Avatar in the header menu](https://linear.app/mysorbet/issue/SRBT-150/hamburger-avatar-in-the-header-menu)
- Complete [SRBT-151: Show header on all pages including when viewing other profiles](https://linear.app/mysorbet/issue/SRBT-151/show-header-on-all-pages-including-when-viewing-other-profiles)
- Move sidebar into header and remove from other pages
- Reveal handle claim UI

Extra changes:
- Colocate notifications code
- Fix "Hire me" being enabled on your own profile
- Remove unused header components
- Remove naive patterns from profile page
- Clean up sidebar imports
- Clean up profile imports

# Screenshots

https://github.com/user-attachments/assets/517d3f84-6248-42c3-a243-1b68ab38374c


src/app/page.tsx
```diff
@@ -1,19 +1,20 @@
...
 import { useAppSelector } from '@/redux/hook';
-import { useEffect, useState } from 'react';
 
 const Home = () => {
   const { user: authUser } = useAuth();
   const reduxUser = useAppSelector((state) => state.userReducer.user);
   const [user, setUser] = useState(authUser || reduxUser);
-  const { toggleOpenSidebar } = useAppSelector((state) => state.userReducer);
+
+  // TODO: Should be in global state, not props.
```
ChristopherPark3, line 17: Are you thinking of implementing redux for this? Bc i know useContext rerenders every child component whenever state updates, and this may actually be what we want because we want to render a different layout for mobile

ChristopherPark3 (approved):
Awesome!

src/app/page.tsx
Cutaiar: This is using redux currently. Just copied the way the sidebar was being opened elsewhere. The header is the only component managing the sidebar's open state so really this could just be a boolean state in header. But I'm not going to change this for now. I'm not convinced we even need redux. If we remove it, I'll change this then :)

PR #55, opened 2024-08-22
Feature: adds location to user profile header

ChristopherPark3:
- Adds location to profile header of a user's profile
- Also slightly tweaks the spacing of each profile header section to make it more compact and waste less space
- Adds minor tweaks to match closer to the figma design

# Before
Screenshot 2024-08-22 at 2 21 12 PM

# After
Screenshot 2024-08-22 at 2 22 16 PM


Cutaiar:
Please verify the new spacing matches design.

Cutaiar (approved):

PR #57, opened 2024-08-23
Claim my sorbet functionality

ChristopherPark3:
- Adds redirect functionality on the 404 page
- Removes the `header` component if a user is not signed in
- Adds a pulsing animation to CTA. Definitely not permanent, just while the design for the page is in the works.

- Was thinking about adding a `Return to my profile` button in the case that a user is already signed in. (Something for the design team, just writing my thoughts down)
# Screenshots

https://github.com/user-attachments/assets/288a7de1-2d4f-4780-a772-b7fa479dc23d


src/app/[username]/page.tsx
```diff
@@ -60,52 +62,74 @@ const ProfilePage = ({ params }: { params: { username: string } }) => {
   const disableHireMe = params.username === user?.accountId.split('.')[0];
   const freelancerFullName = `${freelancer?.firstName} ${freelancer?.lastName}`;
 
+  const handleClaimMyProfile = () => {
+    if (user) {
+      logout();
```
Comment (Cutaiar) [inline src/app/[username]/page.tsx:67]: How come you need to log the user out? As a logged in sorbet user, I wouldn't expect to get logged out by visiting a non-existent link?

src/app/[username]/page.tsx
```diff
@@ -60,52 +62,74 @@ const ProfilePage = ({ params }: { params: { username: string } }) => {
...
-      {isError && <ClaimYourProfile username={params.username} />}
     </>
   );
 };
 
 export default ProfilePage;
 
 /** Local component to display a "Claim your profile CTA when visiting a profile that does not exist" */
-const ClaimYourProfile = (props: { username: string }) => {
+const ClaimYourProfile = (props: {
```
Comment (Cutaiar) [inline src/app/[username]/page.tsx:109]: While you're at it, could you pull this out into a separate component? It has enough functionality now that it warents it I think.

src/app/[username]/page.tsx
```diff
@@ -60,52 +62,74 @@ const ProfilePage = ({ params }: { params: { username: string } }) => {
...
+  username: string;
```
Comment (Cutaiar) [inline src/app/[username]/page.tsx:110]:
```suggestion
  handle: string;
```
I think I want to move in the direction of calling this a handle. What do you think?

Cutaiar (approved):

src/app/[username]/page.tsx
Comment (ChristopherPark3) [reply src/app/[username]/page.tsx:68]: Ahh I was thinking this was for when a user wants to claim that handle, but already has an account. Thinking about it now, I think having a standard 'account' page where the user can change their username would be a lot better.

PR #63, opened 2024-09-03
Updates instagram widget SRBT-140

ChristopherPark3:
- Implements UI for private instagram accounts and instagram accounts that are valid, but haven't posted any pictures.
- Updates `handleWidgetAdd` function in `widget-container` by removing the try catch and allowing React Query to handle the error situation with the `onError` callback function.
  - Before, it was the same toast for each and every widget failure. Now, in `useGetWidgetContent.ts`, the error message that it receives is the error message that is toasted. We just need to make sure that each function that is called in `useGetWidgetContent` is throwing a descriptive message.
- Also fixes the 4x2 media being too small (simple fix from `max-h-full` to `h-full`)

# Screenshots

https://github.com/user-attachments/assets/bb2853b1-eaa3-4882-99c6-d13c80bd166a

Screenshot 2024-09-03 at 1 57 47 AM

### DIFF
--- a/src/components/profile/widgets/widget-container.tsx
+++ b/src/components/profile/widgets/widget-container.tsx
@@ -164,6 +164,10 @@ export const WidgetContainer: React.FC<WidgetContainerProps> = ({
 
       // We're adding a widget that is not a photo
       // So try to fetch required content from the url and persist that content in the database as a new widget
+      /**
+       * Any function that is called by getWidgetContent should throw a descriptive error upon failure using the RQ onError callback
+       * i.e. if we are calling getInstagramProfileMetadata, it should throw a descriptive error that would bubble up to the onError callback
+       */
       const widget = await getWidgetContent({ url: widgetUrl, type });
 
       // If you got the content, add the widget to the layout state and persist it in the database
--- a/src/components/profile/widgets/widget-container.tsx
+++ b/src/components/profile/widgets/widget-container.tsx
@@ -164,16 +164,7 @@ export const WidgetContainer: React.FC<WidgetContainerProps> = ({
 
       // We're adding a widget that is not a photo
       // So try to fetch required content from the url and persist that content in the database as a new widget
-      let widget: any;
-      try {
-        widget = await getWidgetContent({ url: widgetUrl, type });
-      } catch (error) {
-        toast({
-          title: 'Failed to add widget',
-          description: 'If the issue persists, contact support',
-        });
-        return;
-      }
+      const widget = await getWidgetContent({ url: widgetUrl, type });
 
       // If you got the content, add the widget to the layout state and persist it in the database
       const widgetToAdd: ExtendedWidgetLayout = {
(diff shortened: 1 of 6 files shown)


src/components/profile/widgets/widget-instagram.tsx
```diff
@@ -50,65 +47,109 @@ export const InstagramWidget: React.FC<InstagramWidgetType> = ({
...
+          <div className='flex w-2/6 flex-col gap-2'>
+            <WidgetIcon type='InstagramProfile' className='m-0' />
             <div>{localHeader}</div>
           </div>
-          <div className='w-4/6 h-full w-full relative rounded-xl overflow-hidden bg-white text-black'>
-            <div className='grid grid-cols-2 gap-2 w-full h-full'>
-              <ImageGallery images={content.images} heightClass='h-44' />
-              <ImageOverlay />
-            </div>
+          <div className='relative h-full w-full overflow-hidden rounded-xl bg-white text-black'>
```
Cutaiar, line 83: This code is repeated 4 times. Consider a a file scope component below to deduplicate it.

src/hooks/widgets/useGetWidgetContent.ts
```diff
@@ -22,7 +22,7 @@ export const useGetWidgetContent = () => {
       }
     },
     onError: (error) => {
-      toast({ title: 'Error', description: 'Failed to fetch widget content' });
+      toast({ title: 'Oops!', description: error.message });
```
Cutaiar, line 25: Let's use "Something went wrong" rather than oops?

src/lib/service/widgets.ts
```diff
@@ -316,10 +316,12 @@ export const getInstagramProfileMetadata = async ({
...
     return response;
-  } catch (error: any) {
-    throw new Error(
-      `Failed to get Instagram profile metadata: ${error.response.data.message}`
-    );
+  } catch (error: unknown) {
+    if (axios.isAxiosError(error)) {
+      throw new Error(error.response?.data.message);
+    } else {
+      throw new Error('Failed to get Instagram profile metadata');
```
Cutaiar, line 323: Try to include the error from the catch block here. Otherwise, we are losing information

src/components/profile/widgets/widget-container.tsx
```diff
@@ -164,16 +164,7 @@ export const WidgetContainer: React.FC<WidgetContainerProps> = ({
...
-      try {
-        widget = await getWidgetContent({ url: widgetUrl, type });
-      } catch (error) {
-        toast({
-          title: 'Failed to add widget',
-          description: 'If the issue persists, contact support',
-        });
-        return;
-      }
+      const widget = await getWidgetContent({ url: widgetUrl, type });
```
Cutaiar, line 167: Consider adding a comment here and in getWidgetContent describing what you talk about in the PR description about how everything should throw descriptively.

Cutaiar (approved):

PR #64, opened 2024-09-04
Update readme and add gcloud env scripts

Cutaiar:
- Update readme with more professional info
- Update dev instructions to be accurate
- Add info about new env worflow

- Add gcloud env scripts to push and pull local env to gcloud


ChristopherPark3 (approved):
Awesome stuff

PR #65, opened 2024-09-05
Adds animation on resize for share profile dialog

ChristopherPark3:
- Adds `react-use-measure` package that calculates height of a component
- Implements framer motion to animate height changes
- Changes `customDialogClose: string` prop to `hideDefaultCloseButton: boolean` in shadcn `dialog` component for better readability.

** Opted not to add the horizontal animations because I felt that it would be too much and very distracting. Felt that opacity and height resize animation is clean and sleek. Open to discussion tho!

If it were something like this: https://ui.lukacho.com/components/dropdown-menu where a user has a sense of direction because everything is displayed, then i think a horizontal slide makes sense. I digress, but these are just my thoughts.

# Screenshots

https://github.com/user-attachments/assets/3d3b3ebc-c809-4345-9753-ae5910eb8a60

### DIFF
--- a/src/components/profile/share/share-profile-dialog.tsx
+++ b/src/components/profile/share/share-profile-dialog.tsx
@@ -1,6 +1,8 @@
 'use client';
 
+import { AnimatePresence, motion } from 'framer-motion';
 import { Dispatch, SetStateAction, useState } from 'react';
+import useMeasure from 'react-use-measure';
 
 import { Dialog, DialogContent, DialogTrigger } from '@/components/ui/dialog';
 
@@ -45,6 +47,10 @@ export const ShareProfileDialog = ({
     navigator.clipboard.writeText(url);
   };
 
+  const [contentRef, { height: contentHeight }] = useMeasure();
+
+  console.log(contentHeight);
+
   return (
     <Dialog>
       <DialogTrigger asChild>{trigger}</DialogTrigger>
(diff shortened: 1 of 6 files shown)


src/components/ui/dialog.tsx
```diff
@@ -30,14 +30,14 @@ const DialogOverlay = React.forwardRef<
 DialogOverlay.displayName = DialogPrimitive.Overlay.displayName
 
 interface DialogContentProps extends React.ComponentPropsWithoutRef<typeof DialogPrimitive.Content> {
-  customDialogClose?: string;
+  hideDefaultCloseButton?: boolean;
```
Cutaiar, line 33: Nice, I like this better.

src/components/profile/share/share-profile-dialog.tsx
```diff
@@ -45,53 +47,86 @@ export const ShareProfileDialog = ({
...
-            setActive={setActive}
-            handleUrlToClipboard={handleUrlToClipboard}
-          />
-        )}
+        <motion.div
+          // TODO: address this hacky solution for the first view. For some reason, the height is not being calculated correctly when the dialog is opened.
+          animate={{
+            height: activeView === 'ShareYourProfile' ? '372px' : contentHeight,
+          }}
+          className='overflow-hidden'
```
Cutaiar, line 65: What is overflow hidden for?

ChristopherPark3: ahh this is for when the content switches, while the dialog is resizing, if the content is bigger than the previous content, it will overflow outside. 

This makes it so that everything is contained during resizing

src/components/profile/share/share-profile-dialog.tsx
```diff
@@ -45,53 +47,86 @@ export const ShareProfileDialog = ({
...
+            )}
+          </div>
+        </motion.div>
       </DialogContent>
     </Dialog>
   );
 };
+
+/** All this does is animate the opacity of each component and adds a delay for better timing */
+const AnimatedContainer = ({ children }: { children: React.ReactNode }) => {
```
Cutaiar, line 126:
```suggestion
const FadeIn = ({ children }: { children: React.ReactNode }) => {
```

imo, "Container" is almost never a helpful name for things becuase it doesn't say much about what the compnent does. Consdier using a name that describes what it does.

Also, I think you can acheive the same thing with `className='animate animate-in fade-in-10'`

Cutaiar (approved):

src/components/profile/share/share-profile-dialog.tsx
ChristopherPark3: Gotcha let me try it out now!

PR #67, opened 2024-09-05
Wallet Page Base Migration

wonderboy25:
# Wallet Page Base Migration
- Show wallet balance on wallet page
- stop backend call for transaction
- add NEXT_PUBLIC_BASE_RPC_URL on env


src/app/wallet/wallet-container.tsx
```diff
@@ -12,11 +12,21 @@ import { FundsFlow } from '@/app/wallet/funds-flow';
...
 import { WalletBalance } from '@/app/wallet/wallet-balance';
 import { Header } from '@/components/header';
-import { useAuth } from '@/hooks';
+import { useAuth, useWalletBalances } from '@/hooks';
 import { Balances, Transaction, Transactions } from '@/types/transactions';
 
 export const WalletContainer = () => {
-  const { user } = useAuth();
+  const [walletAddress, setWalletAddress] = useState<string>('');
+  // const { user } = useAuth();
```
Cutaiar, line 20: Remove commented code if you do not need it.

src/app/wallet/wallet-container.tsx
```diff
@@ -12,11 +12,21 @@ import { FundsFlow } from '@/app/wallet/funds-flow';
...
+  const {
+    ethBalance,
+    usdcBalance,
+    loading: balanceLoading,
+  } = useWalletBalances(walletAddress);
+  // console.log('user', user);
```
Cutaiar, line 26: Remove commented code if you do not need it.

src/app/wallet/wallet-container.tsx
```diff
@@ -99,12 +117,12 @@ export const WalletContainer = () => {
               isLoading={loading}
               title='Money Out'
               balance={transactions.total_money_out}
-              icon={<Send size={16} />}
+              icon={<MoveUp size={16} />}
```
Cutaiar, line 120: Please attach a screenshot to the PR so that we can see what this change looks like!

src/hooks/useWeb3React.tsx
Cutaiar, line 1: A file containing a single exported hook should be named the same as the hook. Please call this file: `useWalletBalances.ts`. Also note, the `ts` extension rather than `tsx` since there is no JSX in this file.

src/hooks/useWeb3React.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+export const useWalletBalances = (walletAddress: string) => {
+  const [ethBalance, setEthBalance] = useState<string>('0');
+  const [usdcBalance, setUsdcBalance] = useState<string>('0');
+  const [loading, setLoading] = useState(true);
+
+  useEffect(() => {
+    const fetchBalances = async () => {
+      try {
+        if (!walletAddress) {
```
Cutaiar, line 14: I assume this is to check for empty string (`''`). If so, you can be more descriptive and concise with:

```suggestion
        // Ignore empty string wallet address
        if (!walletAddress) return;
```

If you never expect this to be empty string and you are instead expecting and checking for `null` / `undefined`. Please type the hook parameter correctly: `(walletAddress?: string)`.

src/hooks/useWeb3React.tsx
```diff
@@ -0,0 +1,47 @@
...
+        );
+        setUsdcBalance(usdcBalanceFormatted);
+      } catch (error) {
+        console.error('Error fetching wallet balances:', error);
+      } finally {
+        setLoading(false);
+      }
+    };
+
+    fetchBalances();
```
Cutaiar, line 43: You can use an IIFE for this kind of effect to avoid naming a function you only call once.
https://developer.mozilla.org/en-US/docs/Glossary/IIFE#execute_an_async_function

src/lib/config.ts
```diff
@@ -22,6 +23,7 @@ export const config: AppConfig = appConfigSchema.parse({
   googleMapKey: process.env.NEXT_PUBLIC_GOOGLE_MAP_KEY,
   sendbirdAppId: process.env.NEXT_PUBLIC_SEND_BIRD_APP_ID,
   privyAppId: process.env.NEXT_PUBLIC_PRIVY_APP_ID,
+  baseRpcUrl: process.env.NEXT_PUBLIC_BASE_RPC_URL,
```
Cutaiar, line 26: Please pull latest main and merge it. You will see my [latest PR](https://github.com/MySorbet/sorbet-app/pull/68) changes this to an object called `env` which you should add this variable to.

src/app/wallet/wallet-container.tsx
```diff
@@ -31,25 +41,35 @@ export const WalletContainer = () => {
...
+  //       setBalances(response.data.balances);
+  //     }
+  //     setLoading(false);
+  //   }
+  // };
+
+  useEffect(() => {
+    console.log('wallets', wallets);
+    if (wallets) {
+      const wallet = getEmbeddedConnectedWallet(wallets);
```
Cutaiar, line 59: Why do you have to use `getEmbeddedConnectedWallet`. Why can't you just use the first wallet from `useWallets`?

CleanShot 2024-09-06 at 17 32 28@2x

src/app/wallet/wallet-container.tsx
```diff
@@ -31,25 +41,35 @@ export const WalletContainer = () => {
...
   const [loading, setLoading] = useState<boolean>(false);
 
-  const fetchTransactions = async (last_days = 30) => {
-    if (user) {
-      setLoading(true);
-      const response = await getOverview(last_days);
-      if (response && response.data) {
-        setTransactions(response.data.transactions);
-        setBalances(response.data.balances);
+  // const fetchTransactions = async (last_days = 30) => {
```
Cutaiar, line 44: If you want to leave this commented code in, maybe add a `// TODO: Move transactions to base` above it or something like that? Want to avoid a codebase with dead code lying around.

Cutaiar (changes requested):
Great work, Olec. Thanks for your help here. Left some comments. Once addressed, this should be ready to go in!

Please do update the PR description with a title and a bulleted list of the changes. The description ends up being the merge commit so it is helpful to have good information about the change there.

src/hooks/useWeb3React.tsx
wonderboy25: I am going to make other hook methods related web3, so it is good to keep original file name

src/app/wallet/wallet-container.tsx
wonderboy25: Screenshot 2024-09-07 074620

wonderboy25: attached a screenshot

src/app/wallet/wallet-container.tsx
wonderboy25: I confirmed that first wallet can be different. In some cases, first address is shown as my injected address or sometimes it is shown as emebdded one.
Thus we have to use above function

src/hooks/useWeb3React.tsx
Cutaiar: Ah in this case, please do you through with the name change, but put these new hooks in their own files with the same naming convention. To group them, you might consider putting them under `/hooks/web3/`?

src/app/wallet/wallet-container.tsx
Cutaiar: Makes sense. Though I imagine there could be users who would prefer we use their injected address. We will have to look more into this later but this is fine for now.

wonderboy25:
remake PR

Cutaiar (approved):

PR #68, opened 2024-09-06
Use t3 env to validate env on build

Cutaiar:
[T3 Env](https://env.t3.gg/) allows us to validate our env in much the same way we were doing with zod, but with a number of improvements that will help us in the future and the benefit of validating on build. This makes it impossible to build the app with missing env vars. This will help make it easier for me to get a GCP pipeline built.

**Changes**
- Add t3-oss/env and configure `next.config.mjs` to validate on build. Build will fail if env var is missing.
- Add knock vars to validation
- Remove unused `logger.ts` and `NEXT_PUBLIC_SHOW_LOGGER` env var. Will push to GCP when this goes in.

# Screenshots


ChristopherPark3 (approved):
Love this!

PR #72, opened 2024-09-10
SRBT-177 Disables next button if form errors are present in signup flow.

ChristopherPark3:
- Disables next button in `Step1.tsx` by checking the length of the `error` obj provided by React Hook Form.
    - Every other step, the data is optional so there won't be any form errors

### DIFF
--- a/src/components/onboarding/signup/step1.tsx
+++ b/src/components/onboarding/signup/step1.tsx
@@ -231,6 +231,7 @@ const Step1 = () => {
             <Button
               type='submit'
               className='w-full border-[#7F56D9] bg-[#573DF5] text-[#FFFFFF] shadow-sm shadow-[#1018280D]'
+              disabled={Object.keys(errors).length > 0}
             >
               Next
             </Button>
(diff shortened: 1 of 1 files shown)


src/components/onboarding/signup/step1.tsx
```diff
@@ -231,6 +231,7 @@ const Step1 = () => {
             <Button
               type='submit'
               className='w-full border-[#7F56D9] bg-[#573DF5] text-[#FFFFFF] shadow-sm shadow-[#1018280D]'
+              disabled={Object.keys(errors).length > 0}
```
Cutaiar, line 234: Does this disable it on the first render though? IIRC, I used something like this and had that issue. Because there are no errors when the user hasn't taken action yet. But the form is still invalid. Perhaps you want to incorporate some kind of isDirty check for the disabled condition.

ChristopherPark3: Oh yea I am working on that right now! I am implementing an isDirty check as well!

Cutaiar (approved):

PR #73, opened 2024-09-11
SRBT-148 Adds proper disabled state for save button in profile edit modal

ChristopherPark3:
- Disables `save` button when a user has not made any changes
    - For `location-input` and `skill-input` and profile image, this is done through the `setValue` function by passing `shouldDirty` as the options argument.
- Adds `isImageUpdated` property to zod schema so that we can easily check if the image field is dirty through React Hook Form. (The image is not controlled by RHF, but rather local state instead.)
- Implements an effect to reset the form and update default values only when the form has been successfully submitted.
    - Form wasn't syncing to latest changes on save

** Also fixes step1, only disabling continue button on first render (no dirty fields), and when required fields are filled out (`isValid`)

# Video

https://github.com/user-attachments/assets/10af48da-5ba3-495c-ac53-18729c3407b2

### DIFF
--- a/src/components/profile/location-input.tsx
+++ b/src/components/profile/location-input.tsx
@@ -41,7 +41,6 @@ export const LocationInput = <T extends FieldValues>({
             onChange={(e) => {
               setValue(name, e.target.value as PathValue<T, Path<T>>, {
                 shouldDirty: true,
-                shouldTouch: true,
               });
               handleLocationInputChange(e);
             }}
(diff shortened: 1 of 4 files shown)


src/components/profile/location-input.tsx
```diff
@@ -38,7 +38,12 @@ export const LocationInput = <T extends FieldValues>({
             {...register(name, {
               required: 'Location is required',
             })}
```
ChristopherPark3, line 40: Both `onSelect` and `onChange` need the setValue with shouldDirty arg because key press events make value dirty, and selecting from the dropdown is also makes value dirty.

Cutaiar: Nice!

Cutaiar (approved):

PR #74, opened 2024-09-12
SRBT-214 Changes sorbet.xyz to sorbet.io in share profile

ChristopherPark3:
- Changes `sorbet.xyz` to `sorbet.io` in `share-link.tsx`

# Screenshots

Screenshot 2024-09-12 at 9 56 08 AM

Provide screenshots or videos of the changes made if any (optional)

### DIFF
--- a/src/components/profile/share/components/share-link.tsx
+++ b/src/components/profile/share/components/share-link.tsx
@@ -22,8 +22,8 @@ export const ShareLink = ({
         <Image src={Logo} height={32} width={32} alt='logo' />
         <p className='text-base font-medium text-[#101828]'>
           {username.length > 10
-            ? `mysorbet.xyz/${username.slice(0, 10)}...`
-            : `mysorbet.xyz/${username}`}
+            ? `mysorbet.io/${username.slice(0, 10)}...`
+            : `mysorbet.io/${username}`}
         </p>
       </div>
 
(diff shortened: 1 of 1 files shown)


src/components/profile/share/components/share-link.tsx
```diff
@@ -22,8 +22,8 @@ export const ShareLink = ({
         <Image src={Logo} height={32} width={32} alt='logo' />
         <p className='text-base font-medium text-[#101828]'>
           {username.length > 10
-            ? `mysorbet.xyz/${username.slice(0, 10)}...`
-            : `mysorbet.xyz/${username}`}
+            ? `mysorbet.io/${username.slice(0, 10)}...`
+            : `mysorbet.io/${username}`}
```
Cutaiar, line 26: this is already a client component, so i would rather this just be `window.location.hostname` for display and `window.location.origin` for the actual link`.

src/components/profile/share/components/share-link.tsx
```diff
@@ -22,8 +22,8 @@ export const ShareLink = ({
         <Image src={Logo} height={32} width={32} alt='logo' />
         <p className='text-base font-medium text-[#101828]'>
           {username.length > 10
-            ? `mysorbet.xyz/${username.slice(0, 10)}...`
-            : `mysorbet.xyz/${username}`}
+            ? `mysorbet.io/${username.slice(0, 10)}...`
```
Cutaiar, line 25: slice is a fine solution for now, but a more robust way to handle this would be to use css text truncation properties. No need to change, just letting you know.

Cutaiar (approved):

src/components/profile/share/components/share-link.tsx
ChristopherPark3: Sounds good will write that down!

PR #78, opened 2024-09-13
[SRBT-222] Fix Sidebar Balances

Cutaiar:
# [SRBT-222] Fix Sidebar Balances

- Add useEmbeddedWalletAddress for use in wallet-container and sidebar
- Format eth for 4 decimals places in useWalletBalances
- Make local Balance component with proper display logic in sidebar
- Fix styling of balances

# Screenshots

CleanShot 2024-09-13 at 14 12 35@2x

### DIFF
--- a/src/hooks/useEmbeddedWalletAddress.ts
+++ b/src/hooks/useEmbeddedWalletAddress.ts
@@ -1,0 +1,25 @@
+import { getEmbeddedConnectedWallet, useWallets } from '@privy-io/react-auth';
+import { useEffect, useState } from 'react';
+
+/**
+ * Returns the address of the embedded wallet if it exists, otherwise null.
+ * We need a hook like this because privy useWallets will tell use about browser wallets too.
+ */
+export function useEmbeddedWalletAddress(): string | null {
+  const { wallets } = useWallets();
+  const [address, setAddress] = useState<string | null>(null);
+
+  useEffect(() => {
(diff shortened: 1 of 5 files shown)


src/hooks/useEmbeddedWalletAddress.ts
```diff
@@ -0,0 +1,25 @@
...
+/**
+ * Returns the address of the embedded wallet if it exists, otherwise null.
+ * We need a hook like this because privy useWallets will tell use about browser wallets too.
+ */
+export function useEmbeddedWalletAddress(): string | null {
+  const { wallets } = useWallets();
+  const [address, setAddress] = useState<string | null>(null);
+
+  useEffect(() => {
+    if (!wallets || wallets.length === 0) {
```
ChristopherPark3, line 13: what are we thinking happens here? maybe add an error state to return and render in the `Balances` component no existing wallet? (Maybe this may be an irrelevant edge case since Privy provides an embeded wallet)

ChristopherPark3 (approved):
Awesome stuff

src/hooks/useEmbeddedWalletAddress.ts
Cutaiar: So far this is not a case I was able to repro, but I expect this wallet business to change as I work with it more, so will take a not and keep this is mind.

PR #80, opened 2024-09-16
[SRBT-176] Use next/font inter

Cutaiar:
# Use next/font inter

And remove unused font file

This site has never used inter. This is the canonical way to do it.


ChristopherPark3 (approved):
👍

PR #87, opened 2024-09-19
Install Storybook

Cutaiar:
# Install Storybook

- This is the result of running `npx storybook@latest init`
- Fix circular widget/profile imports (they should be relative, not @ absolute)
- Add a couple of stories
- Use a const array `WidgetTypes`

# Screenshots

CleanShot 2024-09-19 at 11 26 30@2x


ChristopherPark3 (approved):
Looking forward to using SB!

PR #91, opened 2024-09-20
Play Button Update

Cutaiar:
# Play Button Update

Add PlayButton component which matches design and use it in spotify/soundcloud widgets

# Screenshots

CleanShot 2024-09-20 at 11 36 04@2x


ChristopherPark3 (approved):
🔥

PR #101, opened 2024-09-27
Fix layout shift

Cutaiar:
# Fix layout shift

By hiding horizontal scroll on `html` and adding a `fix-modal-layout-shift` utility for special cases


ChristopherPark3 (approved):
sweet!

PR #104, opened 2024-10-01
[SRBT-211] Username update functionality

ChristopherPark3:
# [SRBT-ID] Name of this PR

**Changes**

- a
- b
- c

# Screenshots

Screenshots/videos of the changes made (if any, otherwise delete this section)

### DIFF
--- a/src/components/profile/bio-status.tsx
+++ b/src/components/profile/bio-status.tsx
@@ -1,0 +1,17 @@
+import { MAX_BIO_LENGTH } from '@/constant';
+
+export const BioStatus = ({
+  length,
+  isMax,
+}: {
+  length: number;
+  isMax: boolean;
+}) => {
+  return isMax ? (
+    <p className='animate-in slide-in-from-top-1 fade-in-0 mt-1 text-xs text-red-500'>
+      Max of {MAX_BIO_LENGTH} characters
+    </p>
+  ) : (
+    <p className='mt-1 text-xs text-[#344054]'>{length}/100</p>
+  );
+};
--- a/src/components/profile/profile-edit-modal.tsx
+++ b/src/components/profile/profile-edit-modal.tsx
@@ -26,12 +26,15 @@ import {
 import type { User } from '@/types';
 
 import { LocationInput } from './location-input';
+import { BioStatus } from '@/components/profile/bio-status';
 
 const schema = z.object({
   isImageUpdated: z.boolean(),
   firstName: z.string().min(1, 'First name is required'),
   lastName: z.string().min(1, 'Last name is required'),
-  bio: z.string().max(MAX_BIO_LENGTH, `Bio must be at most ${MAX_BIO_LENGTH} characters`),
+  bio: z
+    .string()
+    .max(MAX_BIO_LENGTH, `Bio must be at most ${MAX_BIO_LENGTH} characters`),
   city: z.string(),
   tags: z.array(z.string()).optional(),
 });
@@ -64,6 +67,7 @@ export const ProfileEditModal: React.FC<ProfileEditModalProps> = ({
     setValue,
     getValues,
     reset,
+    watch,
   } = useForm<FormData>({
     resolver: zodResolver(schema),
     defaultValues: {
@@ -75,6 +79,9 @@ export const ProfileEditModal: React.FC<ProfileEditModalProps> = ({
       tags: user?.tags,
     },
   });
+
+  const bioLength = watch('bio').length;
+  const isMaxBioLength = bioLength > MAX_BIO_LENGTH;
 
   const { isDirty } = useFormState({ control });
 
@@ -320,14 +327,7 @@ export const ProfileEditModal: React.FC<ProfileEditModalProps> = ({
                   {...register('bio', { required: 'Bio is required' })}
                   defaultValue={user?.bio}
                 />
-                <label className='text-sm font-normal text-[#475467]'>
-                  Max 100 characters
-                </label>
-                {errors.bio && (
-                  <p className='mt-1 text-xs text-red-500'>
-                    {errors.bio.message}
-                  </p>
-                )}
+                <BioStatus isMax={isMaxBioLength} length={bioLength} />
               </div>
               <div className='item w-full'>
                 <Controller
--- a/src/components/profile/profile-edit-modal.tsx
+++ b/src/components/profile/profile-edit-modal.tsx
@@ -1,12 +1,18 @@
 import { zodResolver } from '@hookform/resolvers/zod';
 import { useQueryClient } from '@tanstack/react-query';
 import { User01, X } from '@untitled-ui/icons-react';
-import { Loader2 } from 'lucide-react';
+import { CircleAlert, CircleCheck, Loader2 } from 'lucide-react';
 import { ChangeEvent, useEffect, useState } from 'react';
-import { Controller, useFormState } from 'react-hook-form';
+import {
+  Controller,
+  ControllerRenderProps,
+  useFormState,
+} from 'react-hook-form';
 import { useForm } from 'react-hook-form';
 import { z } from 'zod';
 
+import { checkHandleIsAvailable } from '@/api/auth';
+import { BioStatus } from '@/components/profile/bio-status';
 import SkillInput from '@/components/syntax-ui/skill-input';
 import { Avatar, AvatarFallback, AvatarImage } from '@/components/ui/avatar';
 import { Button } from '@/components/ui/button';
@@ -16,6 +22,14 @@ import {
   DialogHeader,
   DialogTitle,
 } from '@/components/ui/dialog';
+import {
+  Form,
+  FormControl,
+  FormField,
+  FormItem,
+  FormLabel,
+  FormMessage,
+} from '@/components/ui/form';
 import { Input } from '@/components/ui/input';
 import { MAX_BIO_LENGTH } from '@/constant';
 import {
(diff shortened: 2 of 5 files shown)


src/components/profile/profile-edit-modal.tsx
```diff
@@ -212,162 +259,195 @@ export const ProfileEditModal: React.FC<ProfileEditModalProps> = ({
...
+                    Create a short bio
+                  </FormLabel>
+                  <FormControl>
+                    <textarea
+                      className='border-1 text-textPlaceholder mt-1 w-full resize-none rounded-lg border border-[#D0D5DD] p-3 text-base font-normal text-[#667085] focus:outline-none'
+                      placeholder='A few words about yourself'
+                      rows={4}
+                      {...form.register('bio', { required: 'Bio is required' })}
+                      defaultValue={user?.bio}
                     />
```
Cutaiar, line 415: Use shad textarea

https://arc.net/l/quote/wtpceesq

src/components/profile/profile-edit-modal.tsx
```diff
@@ -212,162 +259,195 @@ export const ProfileEditModal: React.FC<ProfileEditModalProps> = ({
...
+                <FormItem>
+                  <FormLabel className='text-textSecondary'>
+                    Create a short bio
+                  </FormLabel>
+                  <FormControl>
+                    <textarea
+                      className='border-1 text-textPlaceholder mt-1 w-full resize-none rounded-lg border border-[#D0D5DD] p-3 text-base font-normal text-[#667085] focus:outline-none'
+                      placeholder='A few words about yourself'
+                      rows={4}
+                      {...form.register('bio', { required: 'Bio is required' })}
```
Cutaiar, line 413: Can't you move this required bit to the form validator above? Seems strange to have validation logic and messaging in more than one place.

src/components/profile/profile-edit-modal.tsx
```diff
@@ -212,162 +259,195 @@ export const ProfileEditModal: React.FC<ProfileEditModalProps> = ({
...
+                          handleInputChange(field, e.target.value)
+                        }
+                        className='text-textPlaceholder focus:outline-none focus:ring-0'
+                      />
+                      {errors.handle ? (
+                        <CircleAlert className='absolute right-4 top-3 h-4 w-4 text-[#D92D20]' />
+                      ) : (
+                        <CircleCheck className='absolute right-4 top-3 h-4 w-4 text-[#00A886]' />
+                      )}
+                    </div>
```
Cutaiar, line 333: A better way to resuse this handle input from the signup experience would be to extract it (and its change handler) to it's own component that can be used within a shad form. Similar to waht you did for location input.

src/components/profile/profile-edit-modal.tsx
```diff
@@ -55,16 +55,34 @@ export const ProfileEditModal: React.FC<ProfileEditModalProps> = ({
...
+      .string()
+      .min(1, { message: 'Handle is required' })
+      .max(25, { message: 'Handle must be less than 25 characters' })
+      .regex(/^[a-z0-9-_]+$/, {
+        message:
+          'Handle may only contain lowercase letters, numbers, dashes, and underscores',
+      }) // Enforce lowercase, no spaces, and allow dashes
+      .refine((handle) => refine(handle, user?.handle ?? ''), {
+        message: 'This handle is already taken',
+      }),
```
Cutaiar, line 79: A better way to share the validation logic for handles would be to not just extract the refine fn (as you have already done), but to extract this zod validation itself. Then you would have 

```
handle: handleValidation,
...
```

This could be bundled with the extracted handle input i mention below an the refine function.

src/components/profile/location-input.tsx
```diff
@@ -17,12 +17,14 @@ interface LocationInputProps<T extends FieldValues> {
   name: Path<T>;
   register: UseFormRegister<T>;
   setValue: UseFormSetValue<T>;
+  className?: string
```
Cutaiar, line 20: I would add a little `/** note */` that classname is applied to the internal input element, not the root element.

src/components/profile/bio-status.tsx
```diff
@@ -0,0 +1,17 @@
+import { MAX_BIO_LENGTH } from '@/constant';
+
+export const BioStatus = ({
```
Cutaiar, line 3: I would consier calling this something like `BioMessage` since you use it in a very similar way to shad `FormMessage`

src/app/[handle]/page.tsx
```diff
@@ -63,6 +65,15 @@ const ProfilePage = ({ params }: { params: { handle: string } }) => {
   const hideShare = !isMyProfile || !user;
   const freelancerFullName = `${freelancer?.firstName} ${freelancer?.lastName}`;
 
+  /**
```
Comment (Cutaiar) [inline src/app/[handle]/page.tsx:68]: Love the documentation here.

src/app/[handle]/page.tsx
```diff
@@ -76,7 +87,7 @@ const ProfilePage = ({ params }: { params: { handle: string } }) => {
           {!isPending && freelancer && (
             <>
               <Profile
-                user={freelancer}
+                user={freelancerResponse.data}
```
Comment (Cutaiar) [inline src/app/[handle]/page.tsx:90]: Why is this change needed?

tailwind.config.ts
```diff
@@ -81,6 +81,12 @@ const config = {
           DEFAULT: '#573DF5',
           foreground: '#573DF5',
         },
+        textPlaceholder: {
+          DEFAULT: '#667085',
+        },
+        textSecondary: {
+          DEFAULT: '#344054',
+        },
```
Cutaiar, line 89: Usually, you would use `text-secondary` and placeholder text usually uses `text-muted-foreground`. Why use these instead of updating the existing tokens?

Cutaiar (changes requested):

tailwind.config.ts
Cutaiar: Here is some info on shad colors
https://ui.shadcn.com/docs/theming

ChristopherPark3: I made a note to revisit this after our call with design

src/components/profile/widgets/onboarding-drawer/handle-input.tsx
ChristopherPark3, line 1: This diff is going to look a little weird because `handle-input` was a good name for the component I made, and it happens to be what you named it before

src/components/profile/widgets/onboarding-drawer/handle-input.tsx
```diff
@@ -1,126 +1,65 @@
...
+  const handleChange = (e: ChangeEvent<HTMLInputElement>) => {
+    setValue(name, e.target.value as PathValue<T, Path<T>>, {
+      shouldDirty: true,
+      shouldValidate: true,
+    });
   };
 
-  return `${baseUrl[type]}${handle}`;
+  return (
+    <>
```
Cutaiar, line 45: The relative div around this component is an implementation detail of it. It needs this to position it's check on the right correctly. Because it is an implementation detail, it should be within the component, not something the caller is responsible to wrap it in.

src/components/profile/widgets/onboarding-drawer/handle-input.tsx
```diff
@@ -1,126 +1,65 @@
...
+  error,
+  className,
+  ...props
+}: HandleInputProps<T>) => {
+  const handleChange = (e: ChangeEvent<HTMLInputElement>) => {
+    setValue(name, e.target.value as PathValue<T, Path<T>>, {
+      shouldDirty: true,
+      shouldValidate: true,
+    });
   };
```
Cutaiar, line 42: This component is missing the mask from the existing implementations change handler:

```ts
  // This change handler allows us to mask the input to only allow valid handles to be typed or pasted
  // We still have zod validate the form, but this creates a better user experience by guiding
  // the user to a valid handle rather than erroring out when they "do something wrong"
  const handleInputChange = (
    field: ControllerRenderProps<FormSchema, 'handle'>,
    value: string
  ) => {
    field.onChange(
      value
        .replace(/[^a-zA-Z0-9 _-]/g, '') // Remove invalid characters
        .replace(/\s+/g, '-') // Replace spaces with dashes
        .toLowerCase() // Convert to lowercase
    );
  };
  ```

File: src/components/onboarding/signup/step1.tsx
Comment (Cutaiar) [inline src/components/onboarding/signup/step1.tsx:1]: The handle input in this file should use the new handle-input component as well

File: src/components/profile/widgets/onboarding-drawer/onboard-handle-input.tsx
Comment (Cutaiar) [inline src/components/profile/widgets/onboarding-drawer/onboard-handle-input.tsx:1]: I think a better name for this would be SocalHandleInput

**Review by Cutaiar — CHANGES_REQUESTED** [review]

File: src/components/profile/widgets/onboarding-drawer/social-handle-input.tsx [hunk 12]
```diff
@@ -0,0 +1,128 @@
...
+import { useImperativeHandle, useRef, useState } from 'react';
+import React from 'react';
+
+import { Button } from '@/components/ui/button';
+import { Input } from '@/components/ui/input';
+import { cn } from '@/lib/utils';
+
+import { WidgetIcon, WidgetTypeWithIcon } from '../widget-icon';
+
+interface HandleInputProps extends React.InputHTMLAttributes<HTMLInputElement> {
```
Cutaiar, line 11: Prop should be named SocialHandleInputProps

src/components/profile/validate-handle.ts
Cutaiar, line 1: The content of this file is only relevant to the handle-input. It should be exported from the same file.

ChristopherPark3: you're saying to move this into handle-input?

PR #107, opened 2024-10-03
Transaction polish

Cutaiar:
# Transaction polish 

Cleaning up after the recent PR for transactions on base

- Sort imports for transaction components
- Move transaction calls to their own file in `/api`
- Remove unused and confusing period selector from wallet page
- Disable top up and send
- File ProfileEditModal under Profile

# Screenshots

CleanShot 2024-10-03 at 17 41 03@2x

image


ChristopherPark3 (approved):
looks great!

PR #111, opened 2024-10-14
[SRBT-157] Better QR code for share dialog

audreyostrom:
# [SRBT-157] Better QR code for share dialog

**Changes**

- Added `qr-code-styling` package to have a customized QR code that's more in-line with Sorbet's branding. This was the settled upon design that is stylish, while still consistently scannable.
- Fixed a hydration error that's caused by opening the share dialog (`Warning: In HTML, <button> cannot be a descendant of <button>. This will cause a hydration error.`) by converting removing a nested button in `share-on-socials.tsx`
- Fixed ``Warning: React does not recognize the `hideDefaultCloseButton` prop on a DOM element.`` in the dialog component by destructuring hideDefaultCloseButton from the props in the DialogContent component.

# Screenshots
Screenshot 2024-10-14 at 12 33 10 PM


src/components/profile/share/share-profile-dialog.tsx
```diff
@@ -60,7 +60,7 @@ export const ShareProfileDialog = ({
         <motion.div
           // TODO: address this hacky solution for the first view. For some reason, the height is not being calculated correctly when the dialog is opened.
           animate={{
-            height: activeView === 'ShareYourProfile' ? '372px' : contentHeight,
+            height: activeView === 'ShareYourProfile' ? '390px' : contentHeight,
```
Cutaiar, line 63: Why is this needed? Seems like it is adding an extra height transition when the dialog first opens.

src/components/ui/dialog.tsx
```diff
@@ -36,8 +36,8 @@ interface DialogContentProps extends React.ComponentPropsWithoutRef<typeof Dialo
 const DialogContent = React.forwardRef<
   React.ElementRef<typeof DialogPrimitive.Content>,
   DialogContentProps
->(({ className, children, ...props }, ref) => {
-  const {hideDefaultCloseButton} = props
+>(({ className, children, hideDefaultCloseButton, ...props }, ref) => {
+  // const {hideDefaultCloseButton} = props
```
Cutaiar, line 40: Please remove this commented line

src/hooks/profile/useGenerateQRCode.ts
```diff
@@ -1,30 +1,35 @@
...
-  return {
-    svgURL: svgUrl,
-    pngURL: pngUrl,
-    isPending: isSvgURLPending || isPngURLPending,
-  };
+export const useGenerateQRCode = () => {
+  if (typeof window !== 'undefined') {
+    const QRCodeStyling = require('qr-code-styling');
+    const qrOptions = {
+      width: 300,
```
Cutaiar, line 5: Why not `import QRCodeStyling, { type Options } from 'qr-code-styling';` at the top?

Cutaiar, line 5:
```suggestion
    const qrOptions: Options = {
```

to type this object as the constructor expects

src/components/profile/share/views/share-on-socials.tsx
```diff
@@ -18,109 +15,85 @@ export const ShareOnSocials = ({
   setActive,
   handleUrlToClipboard,
 }: ShareOnSocialsProps) => {
-  const url = `${window.location.origin}/${username}`;
+  const [url, setUrl] = useState(`${window.location.origin}/${username}`);
+  const qrCode = useGenerateQRCode();
+  const ref = useRef<HTMLDivElement>(null);
```
Cutaiar, line 20: Rather than use state and effects to keep the qr code updated with the url (which never changes). Why not make the url an argument to the `useGenenerateQRCode` hook and move the effect inside (if it is still even needed)? This keeps the implementation details of the QR code generation out of the UI code here.

src/components/profile/share/views/share-on-socials.tsx
```diff
@@ -18,109 +15,85 @@ export const ShareOnSocials = ({
...
+    qrCode?.update({ data: url });
+  }, [url, qrCode]);
 
-    const a = document.createElement('a');
-    a.href = svgURL;
-    a.download = `${username}-sorbet-qrcode.svg`;
-    a.click();
-    a.remove();
-    URL.revokeObjectURL(svgURL);
+  const handleDownload = (fileExt: 'svg' | 'png') => {
```
Cutaiar, line 33: See my comment regarding the `useGenerateQRCode` above -- this could also be a fn returned by the hook to further isolate implementation detials from UI. so something like: 

```ts
const {qrCode, downloadQRCode} = useGenerateQRCode(url)
```

What do you think?

src/components/profile/share/views/share-on-socials.tsx
```diff
@@ -1,13 +1,10 @@
...
-import { useState } from 'react';
-
+import { useEffect, useRef, useState } from 'react';
+import { CheckCircle, Download01 } from '@untitled-ui/icons-react';
 import { Spinner } from '@/components/common';
 import { Button } from '@/components/ui/button';
 import { useGenerateQRCode } from '@/hooks/profile/useGenerateQRCode';
-
 import { Body, Header, ShareLink } from '../components';
 import { ViewProps } from '../share-profile-dialog';
```
Cutaiar, line 9: You can use eslint to autoorganize these imports. Use `cmd .`.

src/components/profile/share/views/share-on-socials.tsx
```diff
@@ -18,109 +15,85 @@ export const ShareOnSocials = ({
   setActive,
   handleUrlToClipboard,
 }: ShareOnSocialsProps) => {
-  const url = `${window.location.origin}/${username}`;
+  const [url, setUrl] = useState(`${window.location.origin}/${username}`);
+  const qrCode = useGenerateQRCode();
```
Cutaiar, line 19: Would consider calling this ref `qrCodeRef` or something similar to describe what it it used for.

src/components/profile/share/views/share-on-socials.tsx
```diff
@@ -18,109 +15,85 @@ export const ShareOnSocials = ({
...
+      <>
+        <Header
+          title='Share on socials'
+          description='Your unique Sorbet QR code that will direct people to your Sorbet profile when scanned'
+          canGoBack={true}
+          navigateToPrevious={() => setActive('ShareYourProfile')}
+        />
+        <Body>
+          <div className='flex min-h-[256px] w-full items-center justify-center'>
+            {!ref ? <Spinner /> : <div ref={ref} />}
```
Cutaiar, line 52: I thought we would be able to make the rendering of the QR code syncronous by using a library rather than an API. From my testing, looks like there is still a loading period. If this is unavoidable, would prefer it load with a shadcn skeleton of the same height (or at least worth a try to see if it looks more polished)
https://ui.shadcn.com/docs/components/skeleton

Cutaiar (changes requested):

Cutaiar:
image
On my machine, the display QR code has artifacts around the edges (looks low res and chunky). Is there a way to render this in higher resolution, as an svg, or another way that fixes this?

Cutaiar:
Nice work Audrey, left some comments. Please message me if you have any questions :)

public/svg/large-logo.svg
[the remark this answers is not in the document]
audreyostrom: I remember when I was testing it, it looks compressed to me. After switching it the `logo.svg` again, I think my eyes might've been the issue in this case since it's an SVG to begin with.

src/hooks/profile/useGenerateQRCode.ts
audreyostrom: Fixed, and now importing is more standard.  Prior to an [update](https://github.com/kozakdenys/qr-code-styling/pull/64) to `qr-code-styling`, [this](https://github.com/kozakdenys/qr-code-styling/issues/38) was the way to use this library with NextJS projects to avoid `Reference Error: self is not defined` errors.

audreyostrom: Fixed, and also adjusted width and height back to 256 like the original QR code.

src/components/profile/share/views/share-on-socials.tsx
audreyostrom: That would make this simpler/nicer, so I'll go ahead and do that!

src/components/profile/share/share-profile-dialog.tsx
audreyostrom: It's related to me changing the `Button` on Share Your Profile to a `div` (to fix a hydration error that I got when I opened the page regarding nested buttons). It messes up the height of the page and causes it to get cut off like so:
Screenshot 2024-10-15 at 10 32 12 AM

It can be fixed and kept to 372 px by styling the div like a button:
Screenshot 2024-10-15 at 2 05 16 PM

But I wasn't sure if styling like this broke any codebase standards there might be, so I changed the height initially to 390px instead. 

I can either ignore the hydration error or do one of the two ways listed – but if there's an better solution let me know! @Cutaiar

audreyostrom:
Regarding the fuzzy picture on your end @Cutaiar, I changed the QR code to be rendered as an SVG instead using an HTML canvas. I think that removed a lot of artifacts on my end when I viewed it.

src/components/profile/share/views/share-on-socials.tsx
audreyostrom: I re-did it so there's a Skeleton component instead. There might be some delay in the QR code actually displaying due to the animations on the page or delay in images loading in general, but I'm unsure if that's entirely removable since testing with other static image elements also had the same delay.

src/styles/globals.css
audreyostrom, line 1: I didn't end up editing this file, but the linting rules applied when I saved the file undoing my changes. Hence, why the spacing is different.

src/components/profile/share/views/share-on-socials.tsx
```diff
@@ -1,15 +1,13 @@
...
-
 import { Spinner } from '@/components/common';
 import { Button } from '@/components/ui/button';
 import { useGenerateQRCode } from '@/hooks/profile/useGenerateQRCode';
-
+import { CheckCircle, Download01 } from '@untitled-ui/icons-react';
+import { useEffect, useRef, useState, useMemo } from 'react';
 import { Body, Header, ShareLink } from '../components';
 import { ViewProps } from '../share-profile-dialog';
+import { Skeleton } from '@/components/ui/skeleton';
```
Cutaiar, line 10:
```suggestion
import { CheckCircle, Download01 } from '@untitled-ui/icons-react';
import { useEffect, useRef, useState } from 'react';

import { Button } from '@/components/ui/button';
import { Skeleton } from '@/components/ui/skeleton';
import { useGenerateQRCode } from '@/hooks/profile/useGenerateQRCode';

import { Body, Header, ShareLink } from '../components';
import { ViewProps } from '../share-profile-dialog';
```

Cutaiar (changes requested):
Here, I just moved state management into the useGenerateQRCode hook and fixed imports by running eslint auto fix.

src/components/profile/share/share-profile-dialog.tsx
Cutaiar: No broken standards :). Thanks for fixing existing errors, I'm mostly concerned about the extra height transition the change added. Looks like it has gone away with this new change so this is good.

Cutaiar:
This is getting close to going in, left a few more suggestions.

src/components/profile/share/views/share-on-socials.tsx
audreyostrom: Sounds good, I just moved everything and linted the files I edited like you suggested.

Cutaiar: Ah I see. In the diff I just moved the qr code state, not the isCopied bits. But whichever you prefer.

Cutaiar (approved):

src/components/profile/share/views/share-on-socials.tsx
audreyostrom: Ah, I think the suggestion preview might've not have displayed that on my end (or maybe I missed it entirely). I brought back the isCopied bits and the downloadQRCode function so `useGenerateQRCode` only generates the QR code.

PR #114, opened 2024-10-16
[SRBT-172] Data Ownership

ChristopherPark3:
# [SRBT-172] Data Ownership

**Changes**

- Marks unused functions to all api
- Updates a few api calls so that we can check if a user has access to the data

### DIFF
--- a/src/api/chat.ts
+++ b/src/api/chat.ts
@@ -15,11 +15,15 @@ const API_URL = env.NEXT_PUBLIC_SORBET_API_URL;
   @returns blobUrl: string - the blob url of the file
   @throws Error: if request fails
 */
-export async function fetchFile(sendbirdUrl: string, type: string) {
+export async function fetchFile(
+  sendbirdUrl: string,
+  type: string,
+  userId: string
+) {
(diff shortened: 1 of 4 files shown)


src/api/chat.ts
```diff
@@ -15,11 +15,15 @@ const API_URL = env.NEXT_PUBLIC_SORBET_API_URL;
...
+export async function fetchFile(
+  sendbirdUrl: string,
+  type: string,
+  userId: string
+) {
   if (sendbirdUrl.includes('blob:')) {
     return sendbirdUrl;
   }
-  const reqBody = { url: sendbirdUrl };
+  const reqBody = { url: sendbirdUrl, userId };
```
Cutaiar, line 26: As we discussed, since this request is sent with the auth header, can't you use that to get the userId rather than sending it in the body?

Cutaiar: If so, I think most the changes in this PR would be reverted.

ChristopherPark3: Yea I was thinking the same. The only things in this Pr would be the unused flags.

PR #119, opened 2024-10-24
[SRBT-249] Set up Telemetry

ChristopherPark3:
# [SRBT-249] Set up Telemetry
**Changes**

- Sets up PostHog session replays for the first minute (can easily change)
- Implements the beginning of a session recording in `all-set.tsx`, the moment the user presses the button to continue to their profile page.
- The recording is then stopped based on an effect in `/[handle]/page.tsx` after a `60 sec` timeout has completed.
  - Noting that when the `/[handle]/page.tsx` component unmounts (i.e. in the event of page navigation), the timeout may be ended as a result of the clean up function. Thus, recordings can be cut short. 
  - This is how it plays out in my head theoretically, but upon testing, things seem to be working fine.
- Installs `posthog-js` and `posthog-node` packages

** Newest Changes ** 
- Moved the effect for managing the session replay timer to `posthog-provider.tsx`
- Added a feature flag for the session replays, disabling them on dev
  - I just don't want to keep doing recordings everytime i make a new account locally.
  - * There is a plan to create multiple environments on PostHog (prod, dev, staging), just like our other services.


src/components/onboarding/signup/all-set.tsx
```diff
@@ -23,6 +24,11 @@ const AllSet = () => {
     return () => clearTimeout(timer);
   }, []);
 
+  const handleContinue = () => {
+    router.push(`/${userData.handle}`)
```
Cutaiar, line 28: Are you sure you don't want to start the recording before router push? I wonder if this creates a race condition where the push finishes before the recording is started?

src/app/posthog-provider.tsx
```diff
@@ -0,0 +1,46 @@
...
+  }, []);
+
+  return (
+    <PostHogProvider client={posthog}>
+      <PostHogIdentityWrapper>{children}</PostHogIdentityWrapper>
+    </PostHogProvider>
+  );
+}
+
+const PostHogIdentityWrapper = ({ children }: { children: ReactNode }) => {
```
Cutaiar, line 28: little tsdoc about what this does for our future selves?

src/app/posthog-provider.tsx
```diff
@@ -0,0 +1,46 @@
...
+import { PostHogProvider } from 'posthog-js/react';
+import { ReactNode, useEffect } from 'react';
+
+import { useAuth } from '@/hooks';
+import { env } from '@/lib/env';
+
+export function PHProvider({ children }: { children: ReactNode }) {
+  useEffect(() => {
+    posthog.init(env.NEXT_PUBLIC_POSTHOG_KEY, {
+      api_host: process.env.NEXT_PUBLIC_POSTHOG_HOST,
```
Cutaiar, line 13: probably want to access this from `@/lib/env` no?

src/app/posthog-page-view.tsx
Cutaiar, line 1: Have you considered making this a hook like `usePostHogPageView` instead? This is essentially a component that encapsulates a few hook calls, useEffect, and returns null.

src/app/[handle]/page.tsx
```diff
@@ -48,6 +49,17 @@ const ProfilePage = ({ params }: { params: { handle: string } }) => {
     },
   });
 
+  useEffect(() => {
```
Comment (Cutaiar) [inline src/app/[handle]/page.tsx:52]: Perhaps a little note above the effect describing what this does?

Cutaiar (approved):
Approved with suggestions 🦔

src/components/onboarding/signup/all-set.tsx
ChristopherPark3: Ahh that's a good point

src/app/posthog-provider.tsx
ChristopherPark3: nice catch thanks!

src/app/posthog-page-view.tsx
ChristopherPark3: yea that's a good point, but i was just following what the PostHog documentation said to do

ChristopherPark3: one reason i see is to keep the layout a server component. if we make it a hook, it'll have to be a client component, no?

src/lib/flags.ts
```diff
@@ -4,4 +4,5 @@ type FeatureFlags = {
 
 export const featureFlags: FeatureFlags = {
   gigs: false,
+  sessionReplay: process.env.NODE_ENV === 'development' ? false : true,
```
Cutaiar, line 7:
```suggestion
  sessionReplay: process.env.NODE_ENV === 'production',
```

Why not just enable it in production? And not staging?

PR #120, opened 2024-10-25
[SRBT-316] Disable gigs

ChristopherPark3:
# [SRBT-316] Disable gigs

**Changes**
- Creates feature flag object to indicate feature statuses
- Removes gigs from sidebar
- Disables redirect from toasts and Knock notification popover (opens when bell icon is clicked)
- Creates a temporary under dev page for gigs

# Screenshots

Screenshot 2024-10-25 at 5 23 15 PM

### DIFF
--- a/src/app/gigs/page.tsx
+++ b/src/app/gigs/page.tsx
@@ -1,4 +1,5 @@
 import type { Metadata } from 'next';
+import Image from 'next/image';
 
 import {
   InstagramIcon,
@@ -27,10 +28,30 @@ export default function GigsPage() {
             </h1>
           </div>
           <div className='flex w-3/4 items-center justify-center gap-4'>
-            <a>
+            <a
+              href='https://mysorbet.xyz'
+              target='_blank'
+              rel='noopener noreferrer'
+            >
+              <Image
+                src='/svg/logo.svg'
+                height={52}
+                width={52}
+                alt='Sorbet logo'
+              />
+            </a>
+            <a
+              href='https://www.instagram.com/mysorbet.xyz/'
+              target='_blank'
+              rel='noopener noreferrer'
+            >
               <InstagramIcon />
             </a>
-            <a>
+            <a
+              href='https://x.com/mysorbetxyz'
+              target='_blank'
+              rel='noopener noreferrer'
+            >
               <TwitterIcon />
             </a>
           </div>
--- a/src/app/gigs/page.tsx
+++ b/src/app/gigs/page.tsx
@@ -1,10 +1,41 @@
-import { GigsContainer } from './gigs-container';
 import type { Metadata } from 'next';
+
+import {
+  InstagramIcon,
+  TwitterIcon,
+} from '@/components/profile/share/components';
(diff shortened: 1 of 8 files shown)


src/app/gigs/page.tsx
```diff
@@ -1,10 +1,64 @@
...
 };
 
 export default function GigsPage() {
-  return <GigsContainer />;
+  return (
+    <>
+      {featureFlags.gigs ? (
+        <GigsContainer />
+      ) : (
+        <div className='bg-sorbet/25 flex h-full w-full flex-col items-center justify-center'>
```
Cutaiar, line 23: Could consider making this "coming back soon page a separate component.

src/components/notifications/notifications.tsx
```diff
@@ -44,9 +45,12 @@ export const Notifications = () => {
             // TODO: i.e: if we put contract id in the url, the backend would populate the 'data' prop with the contract id
             // For now, it is a simple redirect to 'gigs' page. As we have that more solidified, we can then have more detailed/specific redirects
             onNotificationClick={(item) => {
-              console.log('onNotificationClick: ', item.data);
-              // TODO: look into why this works bc this isn't a client component at the moment
-              router.push('/gigs');
+              if (featureFlags.gigs) {
+                console.log('onNotificationClick: ', item.data);
+                // TODO: look into why this works bc this isn't a client component at the moment
```
Cutaiar, line 50: it works because its parent (`<Header/>`) is a client component. This should be a client component too since it uses router

src/components/profile/profile-header.tsx
```diff
@@ -104,13 +105,17 @@ export const ProfileHeader: React.FC<ProfileHeaderProps> = ({
...
             </Button>
           ) : (
-            <Button
-              className='bg-[#573DF5] px-5 text-base'
-              onClick={onHireMeClick}
-              disabled={disableHireMe}
-            >
-              Hire Me
-            </Button>
+            <>
```
Cutaiar, line 108: Isn't this fragment unnecessary? Should be able to remove the frag and the `{ }`?

src/components/sidebar/sidebar.tsx
```diff
@@ -81,7 +82,7 @@ export const Sidebar: React.FC<SidebarProps> = ({ show }) => {
               </div>
             </div>
             <div>
-              <div className='grid grid-cols-3 gap-2'>
+              <div className='grid grid-cols-2 gap-2'>
```
Cutaiar, line 85: This change is strange b/c i think it will break if you enable the gigs feature flag. Since then the grid is expecting 2 cols and has three children. You could try using [auto cols](https://tailwindcss.com/docs/grid-auto-columns)?

Cutaiar (approved):
Nice. Approved with some suggestions.

src/components/sidebar/sidebar.tsx
ChristopherPark3: would it not just create a new row below?

PR #121, opened 2024-10-28
[SRBT-157] Fix: QR Code not loading on Share Profile Dialog

audreyostrom:
# [SRBT-157] Fix: QR Code not loading on Share Profile Dialog

**Changes**

- Updated dependency array so QR code renders properly.

# Screenshots

Screenshots/videos of the changes made (if any, otherwise delete this section)

### DIFF
--- a/src/hooks/profile/useGenerateQRCode.ts
+++ b/src/hooks/profile/useGenerateQRCode.ts
@@ -44,7 +44,7 @@ export const useGenerateQRCode = (url: string) => {
       qrCode.append(qrCodeRef.current);
       setIsLoadingQRCode(false);
     }
-  }, []);
+  }, [qrCodeRef.current]);
 
   return {
     qrCodeRef,


src/hooks/profile/useGenerateQRCode.ts
```diff
@@ -44,7 +44,7 @@ export const useGenerateQRCode = (url: string) => {
       qrCode.append(qrCodeRef.current);
       setIsLoadingQRCode(false);
     }
-  }, []);
+  }, [qrCodeRef.current]);
```
ChristopherPark3, line 47: It looks like this fixes the problem right now, but it may be worth looking into a little bit deeper because when you update a Ref, it doesn't re-render a component, so the effect may not get ran again when the component re-renders. I don't have the answer as to why and need to research myself, but just a note!

ChristopherPark3 (approved):

src/hooks/profile/useGenerateQRCode.ts
Cutaiar: Agreed. If this is working, let's not spend time on it. but there is some good reading [here](https://medium.com/@teh_builder/ref-objects-inside-useeffect-hooks-eb7c15198780).

Cutaiar (approved):

PR #122, opened 2024-10-28
[SRBT-299] Cropping, Link Editing, and Image Replacement for Widgets

audreyostrom:
# [SRBT-299] Cropping, Link Editing, and Image Replacement for Widgets

**Changes**

*Image Cropping*

- Image cropping is now supported for Photo widgets, regardless of image size. The library used was [react-easy-crop](https://www.npmjs.com/package/react-easy-crop). 
- When cropping is active, the user shouldn't be able to resize the widget or preform any other non-cropping actions with it. The widget will maintain its position on the screen, however.
- You cancel out of cropping with the escape button, similar to Bento.
- Demonstration:

Uploading cropping images.mp4…

*Updating Links*
- The user can update links for Link widgets. 
-  Demonstration (for valid and non-valid links):

https://github.com/user-attachments/assets/a83c3cc4-b591-4838-a38d-81ad846d61d9

*Updating Images*
- The user can update images for Link widgets. 
-  Demonstration: 

https://github.com/user-attachments/assets/dcd3010f-ca0c-4c37-afeb-8a37355cc7e5


src/api/widgets.ts
```diff
@@ -93,6 +93,44 @@ export const updateWidget = async (
   }
 };
 
+/** Updating the links of image widgets (only widget's redirectUrl should be updated) */
+export const updateWidgetLink = async (widgetLayoutItem: WidgetLayoutItem) => {
+  // redirectUrl should be set to what we want already in the payload
+  let payload: any = {
```
ChristopherPark3, line 99: I think the 'any' is unnecessary as you don't redefine it anywhere so it can even be a 'const'. Just something to note as well, try to avoid the `any` type as it can kind of hurt in the long run. I'm definitely guilty of it, and just wanted to point it out so you avoid the same mistakes I made :)

src/api/widgets.ts
```diff
@@ -93,6 +93,44 @@ export const updateWidget = async (
...
+  };
+
+  try {
+    const res = await axios.patch(
+      `${env.NEXT_PUBLIC_SORBET_API_URL}/widgets/${widgetLayoutItem.id}`,
+      payload,
+      await withAuthHeader()
+    );
+    return res.data;
+  } catch (error: any) {
```
ChristopherPark3, line 110: Along the same note, the api calls are kind of weird to type as well. You can definitely find it somewhere in the code base, but I believe axios says to cast the error as `unknown`, and then run some conditionals to see if it is an `AxiosError` and go from there. If you look in `auth.ts`, there's a good example. (Ofc this is in the context of making a network call via axios)

Screenshot 2024-11-05 at 11 21 13 AM

src/components/profile/widgets/add-link.tsx
```diff
@@ -0,0 +1,87 @@
+import React, { useState } from 'react';
```
ChristopherPark3, line 1: One small thing, if you could go through and fix the imports to follow the ESLint rules that would be awesome. If you right click or press 'cmd+.', there should be an option to auto sort the imports. If you don't see the ESLint warnings then lmk and I can try to help you get that working!

src/hooks/profile/useWidgetManagement.ts
```diff
@@ -125,6 +138,165 @@ export const useWidgetManagement = ({
     ]
   );
 
+  /** Handles the replacement of display images for Link and Photo Widgets */
+  const handleNewImageAdd = useCallback(
+    async (key: string, image: File) => {
+      setAddingWidget(true);
+      let widgetUrl: string = '';
```
ChristopherPark3, line 145: Minor: ESLint is saying this cast to 'string' is unnecessary

src/hooks/profile/useWidgetManagement.ts
```diff
@@ -125,6 +138,165 @@ export const useWidgetManagement = ({
...
+          const newObj = {
+            ...existingItem,
+            id: existingItem.i, // Replace 'i' with 'id' for payload
+          };
+          await useUpdateWidgetImageAsync(newObj);
+        }
+      } catch (error) {
+        const message =
+          error instanceof Error ? error.message : 'Something went wrong';
+        toast({
```
ChristopherPark3, line 193: Oh and then here, I think we spoke about moving away from the shadcn toasts to the sonner toasts. Not a big deal in this PR, we can go through and change it in it's own separate PR. Just wanted to remind going forward (I get the old way was already implemented before you touched this file).

src/hooks/profile/useWidgetManagement.ts
```diff
@@ -125,6 +138,165 @@ export const useWidgetManagement = ({
...
+              break;
+
+            default:
+              break;
+          }
+          const newObj = {
+            ...existingItem,
+            id: existingItem.i, // Replace 'i' with 'id' for payload
+          };
+          await useUpdateWidgetImageAsync(newObj);
```
ChristopherPark3, line 188: Minor: theres an error saying that you can't call hook inside of a hook. If you just get rid of 'use' in the name, it'll go away. useUpdateWidgetImageAsync -> updateWidgetImageAsync.

ChristopherPark3: Love the thoughts of optimization!

ChristopherPark3 (approved):
Awesome stuff Audrey!

src/components/profile/widgets/add-link.tsx
audreyostrom: Sorry about that, I forgot to lint everything before I converted this to a PR.

public/svg/section-title.svg
Cutaiar, line 1: Adding a section title icon isn't needed for this PR correct?

src/api/widgets.ts
Cutaiar: In catch blocks, you can leave the error untyped. We have TS set up to interpret them as `unkown` automatically. Then you can do what chris says here to handle the error.

src/components/profile/widgets/add-link.tsx
```diff
@@ -0,0 +1,91 @@
...
+export const AddLink = ({
+  value,
+  onChange,
+  setPopoverOpen,
+  popoverOpen,
+  onSubmission,
+}: AddLinkProps) => {
+  const [isValid, setIsValid] = useState(true); // State to track validity
+
+  const { toast } = useToast();
```
Cutaiar, line 30: Let's use [Sonner](https://ui.shadcn.com/docs/components/sonner) for toasts moving forward.

src/components/profile/widgets/widget-section.tsx
Cutaiar, line 1: Weren't you excluding section title from this PR?

src/types/widgets.ts
```diff
@@ -143,13 +144,14 @@ export const WidgetTypes = [
...
 
-export type WidgetSize = 'A' | 'B' | 'C' | 'D';
+export type WidgetSize = 'A' | 'B' | 'C' | 'D' | 'Section';
 
 export const WidgetDimensions: Record<WidgetSize, { w: number; h: number }> = {
   A: { w: 2, h: 2 },
   B: { w: 4, h: 4 },
   C: { w: 4, h: 2 },
   D: { w: 2, h: 4 },
+  Section: { w: 4, h: 1 },
```
Cutaiar, line 154: Similar to other comments, I though section titles was excluded from this PR?

Cutaiar (changes requested):
This is just a quick sweep. Will take a more detailed look later today.

src/api/widgets.ts
audreyostrom: Thank you@

src/components/profile/widgets/widget-section.tsx
audreyostrom: Re: this and the other comments, I'll remove any reference to sections.

src/api/widgets.ts
```diff
@@ -54,10 +67,16 @@ export const updateWidgetsBulk = async (
       await withAuthHeader()
     );
     return response;
-  } catch (error: any) {
-    throw new Error(
-      `Failed to update widgets in bulk: ${error.response.data.message}`
-    );
+  } catch (error: unknown) {
```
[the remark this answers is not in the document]
Cutaiar: Same for the rest of the caught errors in this file

src/hooks/profile/useWidgetManagement.ts
```diff
@@ -125,6 +134,161 @@ export const useWidgetManagement = ({
...
+    async (key: string, url: string) => {
+      try {
+        const existingItem = layout.find((item) => item.i === key);
+        if (existingItem) {
+          existingItem.redirectUrl = url;
+          const newObj = {
+            ...existingItem,
+            id: existingItem.i, // Replace 'i' with 'id' for patch endpoint
+          };
+          await updateWidgetLinkAsync(newObj);
```
Cutaiar, line 276: Since all you are updating about the widget is it's link, I think it's more appropriate for for `updateWidgetLinkAsync` to just take the link (and widget id) as a param, rather requiring the caller to construct a whole widget object just to update the link. You can propgate this change all the way down. so `updateWidgetLink` from the api folder can get this change too.

Seems like the same change could apply to `updateWidgetImage` as well. Please consider this -- if it turns out its not cleaner, that's fine, but generally, functions should take as params only what they need.

src/hooks/profile/useWidgetManagement.ts
```diff
@@ -125,6 +134,161 @@ export const useWidgetManagement = ({
...
+        }
+      } catch (error) {
+        const message =
+          error instanceof Error ? error.message : 'Something went wrong';
+        toast(`We couldn't update a widget`, {
+          description: message,
+        });
+      }
+    },
+    [layout]
```
Cutaiar, line 288: Eslint is complaining about this dep array. Are you getting these warnings?

Cutaiar:
As discussed, we probably don't want to allow zoom cropping. So the scrolling when cropping should not have an effect. Images that are longer than they are tall should only be able to be moved horizontally. Images that are taller than they are long should only be able to be moved vertically. And square images should probably have cropping disabled. Like bento:

CleanShot 2024-11-07 at 12 29 01@2x

Please make an account and play with bento.me to get an idea of their UX.

Cutaiar:
The location and size of link dropdown should match design:
CleanShot 2024-11-07 at 12 34 46@2x

design:

CleanShot 2024-11-07 at 12 35 34@2x

Cutaiar:
When cropping, we don't need this grey background. Just the black border and image. (as per design and bento)

Cutaiar:
When cropping, clicking outside the widget should apply the cropping. As it stands, I couldn't figure out how to apply the cropping until i hovered the widget again and saw the green crop button.

src/components/ui/input.tsx
```diff
@@ -19,8 +20,9 @@ const Input = React.forwardRef<HTMLInputElement, InputProps>(
         <input
           type={type}
           className={cn(
-            'flex h-10 w-full rounded-md border border-input bg-background px-3 py-2 text-sm ring-offset-background file:border-0 file:bg-transparent file:text-sm file:font-medium placeholder:text-muted-foreground focus:outline-none focus:ring-2 focus:ring-ring focus:ring-offset-2 disabled:cursor-not-allowed disabled:opacity-50',
+            'flex h-10 w-full rounded-md border border-input bg-background px-3 py-2 text-sm ring-offset-background file:border-0 file:bg-transparent file:text-sm file:font-medium placeholder:text-muted-foreground focus:outline-none disabled:cursor-not-allowed disabled:opacity-50',
             suffix && 'pr-10',
+            noRing === undefined && noRing === true && 'focus:ring-2 focus:ring-ring focus:ring-offset-2',
             prefix && 'rounded-l-none',
```
Cutaiar, line 26: Hm. Could you comment on what this line is doing? When would `noRing` be both `undefined` and `true`?

src/components/profile/widgets/widget.tsx
```diff
@@ -91,7 +115,6 @@ export const Widget: React.FC<WidgetProps> = ({
         }
       }
     }
-    // TODO: Maybe widgets should be anchors?
   };
```
Cutaiar, line 118: This todo should not be removed. I still beleive widgets should be anchors rather than divs, but this will take a some effort to solve.

Cutaiar:
Something in this PR makes it so that any change to a widget (uploading an image to a link widget, adding a new widget, etc) causes all the widgets to go into a loading state. This should not be the case. Updating or adding a widget should not effect the others visually.

src/api/widgets.ts
```diff
@@ -100,8 +179,14 @@ export const deleteWidget = async (id: string) => {
...
-  } catch (error: any) {
-    throw new Error(`Failed to delete widget: ${error.response.data.message}`);
+  } catch (error: unknown) {
+    if (axios.isAxiosError(error)) {
+      throw new Error(
+        `Axios error: Failed to delete widget: ${error.response?.data.message}`
+      );
+    } else {
+      throw new Error(`Non-axios error: Failed to delete widget: ${error}`);
+    }
```
Cutaiar, line 189: Rather than repeating this in every catch block, you could consider extracting this logic to a function like `catchAndRethrowWidgetErrror(error: unkown, message?: string)` or something like this. that way, this code is less redundant, and upgrades to error handling can benefit all these functions.

Cutaiar (changes requested):
Made some individual comments, and grouped some into a review here. Great work and a good start to this pretty large feature. It is coming along but needs some polish and aligment with design.

src/components/ui/input.tsx
audreyostrom: That line has been more appropriately configured to make sense because you are right. I also just refactored it to be a variant as well since I apply similar styling on the Section Title branch.

src/hooks/profile/useWidgetManagement.ts
audreyostrom: Yes, the ones relevant to this PR are fixed.

src/hooks/profile/useWidgetManagement.ts
audreyostrom: I applied this change for widget link. For updateWidgetImage, I more aptly renamed it to to `updateWidgetContent` to account for the fact that the name for the image URL for all the widgetContentTypes isn't consistent (i.e. for photo widgets it's `image` and for links it's `heroImageUrl`)

Cutaiar (approved):

public/svg/empty-widget-background.svg
```diff
@@ -0,0 +1,9 @@
+<svg width="260" height="180" viewBox="0 0 260 180" fill="none" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
```
Cutaiar, line 1: Seems like this svg should be removed now.

src/components/profile/widgets/widget-photo.tsx
```diff
@@ -10,12 +10,46 @@ interface PhotoWidgetType {
...
         src={content.image}
         alt='Photo content'
-        className='relative h-full w-full object-cover'
+        className={
+          isCropped && croppedArea
+            ? 'absolute'
+            : `relative h-full w-full object-cover`
+        }
+        style={isCropped && croppedArea ? calculateStyles(croppedArea) : {}}
       />
```
Cutaiar, line 53:
```suggestion
        style={isCropped && croppedArea ? calculateStyles(croppedArea) : undefined}
```

src/components/profile/widgets/widget-photo.tsx
```diff
@@ -10,12 +10,46 @@ interface PhotoWidgetType {
...
 
 export const PhotoWidget: React.FC<PhotoWidgetType> = ({ content }) => {
+  // stylings for cropped images derived from official react-easy-crop examples
+  // reference: https://codesandbox.io/p/sandbox/react-easy-crop-with-live-output-kkqj0?file=%2Fsrc%2FApp.jsx%3A49%2C11
+  // this is the only approach that seems to handle images of any dimensions.
+
+  const croppedArea =
+    (content as PhotoWidgetContentType).croppedArea ?? undefined;
+  const isCropped = (content as PhotoWidgetContentType).isCropped;
+
```
Cutaiar, line 20:
```suggestion
  const croppedArea = content.croppedArea;
  const isCropped = content.isCropped;
```

Doesn't look like this cast is needed

src/components/profile/widgets/modify-widget-image.tsx
```diff
@@ -0,0 +1,176 @@
...
+  identifier: string;
+  hasImage: boolean;
+  addImage: (key: string, image: File) => Promise<void>;
+  removeImage: (key: string) => Promise<void>;
+  setErrorInvalidImage: Dispatch<SetStateAction<boolean>>;
+  restoreImage?: (event: React.MouseEvent<HTMLDivElement>) => void;
+  className?: string;
+}
+
+export const ModifyImageWidget: React.FC<ModifyImageWidgetProps> = ({
```
Cutaiar, line 24: A TSDoc comment describing what this component is would be nice. Also seems like a more appropriate name would be "ModifyImageControls". We already use the term widget for the components on the profile, best not to overload it.

src/components/profile/widgets/modify-widget-image.tsx
```diff
@@ -0,0 +1,176 @@
...
+
+interface LinkedPictureIconProps {
+  className?: string;
+  width?: number | string;
+  height?: number | string;
+}
+
+const LinkedPictureIcon: React.FC<LinkedPictureIconProps> = ({
+  className,
+  width = 24,
```
Cutaiar, line 136: You can use `className='size-5'` rather than exposing specific width and height props. This way it behaves like the other icons we use.

src/components/profile/widgets/cropping-widget.tsx
```diff
@@ -0,0 +1,251 @@
...
+            (content as PhotoWidgetContentType).croppedArea
+              ? (content as PhotoWidgetContentType).croppedArea
+              : undefined
+          }
+        />
+        <div
+          className={`absolute left-1/2 -translate-x-1/2 -translate-y-1/2 transform ${
+            isPopoverOpen
+              ? ''
+              : 'opacity-0 transition-opacity duration-300 hover:opacity-100'
```
Cutaiar, line 217:
```suggestion
          className={cn('absolute left-1/2 -translate-x-1/2 -translate-y-1/2 transform', isPopoverOpen && 'opacity-0 transition-opacity duration-300 hover:opacity-100')
```

src/components/profile/widgets/widget-behance.tsx
```diff
@@ -1,21 +1,52 @@
...
 import { ImageOverlay } from './image-overlay';
 import { WidgetDescription } from './widget-description';
 import { WidgetIcon } from './widget-icon';
 
 interface BehanceWidgetType {
+  identifier: string;
+  showControls?: boolean;
+  setErrorInvalidImage: Dispatch<SetStateAction<boolean>>;
+  addImage: (key: string, image: File) => Promise<void>;
+  removeImage: (key: string) => Promise<void>;
```
Cutaiar, line 21: these and the other props you've added to every widget are pretty repetitive. Let's extract them to an interface like `BaseWidgetProps` and have each widget interface extend this interface. This way, you don't have to make changes to almost 20 files if you want to change that interface.

src/components/profile/widgets/widget-dribbble.tsx
```diff
@@ -52,13 +94,21 @@ export const DribbbleWidget: React.FC<DribbbleWidgetType> = ({
...
+          <div className='relative flex-grow'>
+            {showControls && (
+              <ModifyImageWidget
+                hasImage={content.image ? true : false}
+                restoreImage={restoreImage}
+                setErrorInvalidImage={setErrorInvalidImage}
+                identifier={identifier}
+                addImage={addImage}
+                removeImage={removeImage}
+                className='absolute left-1/2 top-0 z-20 flex -translate-x-1/2 -translate-y-1/2 transform items-center opacity-0 transition-opacity group-hover:opacity-100'
```
Cutaiar, line 106: These classes are repeated on every `ModifyImageWidget`. If you wanted to tweak this, you'd have to change it in many files. Would be better to share these classes via a component which wraps `ModifyImageWidget` such as `ModifyImageWidgetWithHover` or something like that.

src/components/profile/widgets/widget-dribbble.tsx
```diff
@@ -52,13 +94,21 @@ export const DribbbleWidget: React.FC<DribbbleWidgetType> = ({
...
-            <img
-              src={content.image}
-              alt='Dribbble content'
-              className='h-full w-full object-cover'
-            />
-            <ImageOverlay />
+          <div className='relative flex-grow'>
+            {showControls && (
+              <ModifyImageWidget
+                hasImage={content.image ? true : false}
```
Cutaiar, line 100:
```suggestion
                hasImage={Boolean(content.image)}
```
or
```suggestion
                hasImage={!!content.image}
```

Cutaiar: And the same for the rest of the places you render `ModifyImageWidget`

src/components/profile/widgets/widget.tsx
```diff
@@ -278,8 +425,9 @@ export const Widget: React.FC<WidgetProps> = ({
           type !== 'SectionTitle' && 'bg-white',
           type === 'SectionTitle' && 'py-4'
         )}
+        id={identifier}
         key={identifier}
-        onClick={onWidgetClick}
+        onClick={!showControls ? onWidgetClick : () => {}} // Don't redirect if editing dashboard, similar to Bento
       >
```
Cutaiar, line 431:
```suggestion
        onClick={!showControls ? onWidgetClick : undefined} // Don't redirect if editing dashboard, similar to Bento
```

src/components/profile/widgets/widget.tsx
```diff
@@ -291,19 +439,34 @@ export const Widget: React.FC<WidgetProps> = ({
...
             )}
           >
             <div className='flex flex-row gap-1'>
               {type !== 'SectionTitle' && (
                 <ResizeWidget
+                  redirectUrl={redirectUrl}
                   onResize={onWidgetResize}
+                  onEditLink={onWidgetLinkEdit}
+                  setPopoverOpen={setIsPopoverOpen}
+                  popoverOpen={isPopoverOpen}
```
Cutaiar, line 461: Is there a good reason that the open state of the add link input needs to be pulled all the way out to this component? 

If so, please give it a more descriptive name like `isAddLinkOpen` so that it is clear what the state is for. 

If not, seems that this state could be local to the `ResizeWidget` component.

src/components/profile/widgets/add-link.tsx
```diff
@@ -0,0 +1,113 @@
...
+        onSubmission();
+      } else if (!isValidUrl(value)) {
+        setIsValid(false);
+        toast('Invalid Link', {
+          description:
+            'Double check the URL you entered redirects to a legitimate website.',
+        });
+        return;
+      } else {
+        // otherwise, update widget's link
```
Cutaiar, line 57: Doesnt look like you are calling onsubmission here?

src/components/profile/widgets/add-link.tsx
```diff
@@ -0,0 +1,113 @@
...
+  popoverOpen: boolean;
+  setPopoverOpen: (open: boolean) => void; // Prop for managing popover state
+}
+
+export const AddLink = ({
+  value,
+  onChange,
+  setPopoverOpen,
+  popoverOpen,
+  onSubmission,
```
Cutaiar, line 27: The interface for this component is a little confusing. Rather than make it a controlled component component, I would let this component hold the state for the link value. Then when a submission happens, you have the onSubmission callback pass the newvalue to the parent. This way, you can remove the value and onChange props. And just have an onSubmit prop to communicate to the parent.

This helps because you should normalize the links that are passed to the parent anyway. DM me and I can help elucidate this more.

src/components/profile/widgets/modify-widget-image.tsx
```diff
@@ -0,0 +1,176 @@
...
+  restoreImage,
+  className,
+}) => {
+  const btnClass = 'h-4 w-7 flex items-center justify-center';
+  const dividerClass = 'h-4 w-[2.5px] bg-[#344054] rounded-full mx-2';
+
+  const handleImageUpload = (e: React.ChangeEvent<HTMLInputElement>) => {
+    const file = e.target.files ? e.target.files[0] : undefined;
+    if (file) {
+      const validExtensions = ['jpg', 'jpeg', 'png', 'gif', 'svg'];
```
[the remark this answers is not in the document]
audreyostrom: Sounds good. I moved this function to the `util.tsx` folder.

src/components/profile/widgets/widget-behance.tsx
audreyostrom: Added BaseWidgetProps to `src/types/widget.ts` and added the appropriate extensions

src/components/profile/widgets/cropping-widget.tsx
audreyostrom: Removed all instances like this and replaced them with more appropriate `cn()` uses

src/components/profile/widgets/add-link.tsx
audreyostrom: Added back.

src/components/profile/widgets/widget.tsx
audreyostrom: Originally, this snippet on main was:

```
{showControls && (
          <div
            className={cn(
              'absolute bottom-0 left-1/2 -translate-x-1/2 transform opacity-0 transition-opacity duration-300 group-hover:opacity-100',
              type !== 'SectionTitle' && 'translate-y-5',
              type === 'SectionTitle' && '-translate-y-1' // to account for margins on the background of section titles
            )}
          >
            <div className='flex flex-row gap-1'>
              {type !== 'SectionTitle' && (
                <ResizeWidget
                  onResize={onWidgetResize}
                  initialSize={initialSize}
                />
              )}
              <Button
                variant='outline'
                size='icon'
                className='rounded-full border-gray-800 bg-gray-800 text-white hover:bg-gray-800 hover:text-white'
                onClick={(e) => {
                  e.stopPropagation();
                  handleRemove(identifier);
                }}
              >
                <Trash2 size={18} />
              </Button>
            </div>
          </div>
        )}
  ```
The reason why I moved this extra logic in `widget.tsx` is to ensure that the popover stays visible when editing a link. When the user starts typing in the input box while editing a link, they're technically no longer hovering over this  div containing the trash button on resize widget – making it disappear.

I renamed it to `isAddLinkOpen` as you suggested.

File: src/components/profile/widgets/widget-dribbble.tsx (hunk 22 above)
Comment (audreyostrom) [reply src/components/profile/widgets/widget-dribbble.tsx]: Reading it closely, all the strings are the same. I just moved this className into the component itself and there were no differences.

File: src/components/profile/widgets/modify-widget-image.tsx [hunk 29]
```diff
@@ -0,0 +1,155 @@
...
+  restoreImage?: (event: React.MouseEvent<HTMLDivElement>) => void;
+  className?: string;
+}
+
+/**
+ * Component that appears above the contents of any given widget to
+ * provide the option to replace the widget's content image, remove it,
+ * or restore it to the website's scraped image.
+ */
+export const ModifyImageControls: React.FC<ModifyImageControlsProps> = ({
```
Cutaiar, line 30: file name should generally be the same as the main exported component, but in kebab case. So this would be `modify-image-controls.tsx`

src/components/profile/widgets/widget-dribbble.tsx
Cutaiar: It doesn't look like there was any change here. The idea with this comment is to reduce redundant code. This string is repeated 21 times. It would be better to write it once and apply it to `ModifyImageWidget` or a wrapping component.

Cutaiar:
Left two more comments. When they are resolved, go ahead and merge!

src/components/profile/widgets/widget-dribbble.tsx
audreyostrom: Sorry I thought you were just referring to the className string being repeated, which I removed from the numerous instances. Am I wrong?

PR #123, opened 2024-11-05
Increases session replay time limit to 5 min

ChristopherPark3:
# Increases session replay time limit to 5 min

- Increases session replay time limit
- Creates a session replay time limit constant in constants.ts


src/constant/constants.ts
```diff
@@ -2,3 +2,6 @@
 export const MAX_BIO_LENGTH = 100;
 export const MAX_NUM_SKILLS = 5;
 export const MAX_CHARS_PER_SKILL = 20;
+
+/** PostHog */
+export const SESSION_REPLAY_DURATION = 5 * 60 * 1000;
```
Cutaiar, line 7:
```suggestion
export const SESSION_REPLAY_DURATION = 5 * 60 * 1000; // 5 minutes
```

Cutaiar (approved):

PR #137, opened 2024-11-19
[SRBT-331] Implement "My Accounts" Card

audreyostrom:
# [SRBT-331] Implement "My Accounts" Card

**Changes**

- Added "My Account Widget" to wallet page. For right now, it only displays USDC balances, but this could be expanded at a later point.

# Screenshots

Screenshot 2024-11-20 at 8 48 35 AM

Screenshot 2024-11-20 at 8 48 28 AM

Screenshot 2024-11-20 at 8 50 59 AM


src/app/wallet/wallet-balance.tsx
```diff
@@ -49,7 +49,7 @@ export const WalletBalance: React.FC<WalletBalanceProps> = ({
           <div>
             <div className='flex items-center gap-2'>
               <span className='rounded-full bg-black p-2 text-white'>
-                <Wallet size={18} />
+                <Wallet03 width={18} height={18} />
```
Cutaiar, line 52: Usually, we have been sizing icons with tailwind for consistency. i.e. `className='size-4'`. This would be 16px rather than 18. To get 18px in a way that responds to the users text size, you can use `size-[1.125rem]`. Ping me if you have questions or want to learn more about why :)

src/app/wallet/my-accounts.tsx
```diff
@@ -0,0 +1,73 @@
...
+    </div>
+  );
+};
+
+const USDCAccountItem: React.FC<{
+  icon: React.ReactNode;
+  balance: string;
+  address: string | null;
+}> = ({ icon, balance, address }) => {
+  const truncatedAddress = `${address?.slice(0, 5)}...${address?.slice(-5)}`;
```
Cutaiar, line 41: Once your other PR goes in, let's share the `truncateAccount` function for all wallet address displays.

audreyostrom:
@Cutaiar @ChristopherPark3 I addressed all the original issues with this branch

src/app/wallet/account-modal.tsx
```diff
@@ -0,0 +1,64 @@
...
+  accountModalVisible,
+  address,
+  handleModalVisible,
+}) => {
+  return (
+    <Dialog open={accountModalVisible} onOpenChange={handleModalVisible}>
+      <DialogContent
+        className='p-8 sm:rounded-[32px]'
+        hideDefaultCloseButton={true}
+        aria-describedby={undefined}
```
ChristopherPark3, line 28: Looks great! One thing in regards to accessibility:

We generally want to have a description for screen readers, and a good example of a way to implement that is in `wallet-send-dialog.tsx` lines `264-268`. We want a description for screen readers, but we want to hide it, thus the Radix `VisuallyHidden` components. 

Now that I am looking at it, I need to go back and add descriptions for the other views, but you get the idea. Then I believe you can remove this `aria-described-by` prop

src/app/wallet/wallet-balance.tsx
```diff
@@ -50,7 +50,7 @@ export const WalletBalance: React.FC<WalletBalanceProps> = ({
           <div>
             <div className='flex items-center gap-2'>
               <span className='rounded-full bg-black p-2 text-white'>
-                <Wallet size={18} />
+                <Wallet03 className='size-[1.125rem]' />
```
ChristopherPark3, line 53: is there a reason why you chose rem instead of pixels?

ChristopherPark3 (approved):
Nice work Audrey!

src/app/wallet/account-modal.tsx
audreyostrom: Oh, that's my oversight. Thank you, I'll come up with a better descriptor.

src/app/wallet/account-modal.tsx
```diff
@@ -0,0 +1,64 @@
...
+  DialogContent,
+  DialogHeader,
+  DialogTitle,
+} from '@/components/ui/dialog';
+
+import { CopyIconButton } from '@/components/common/copy-button/copy-icon-button';
+import { Button } from '@/components/ui/button';
+
+interface AccountModalProps {
+  accountModalVisible: boolean;
```
Cutaiar, line 13: Reccomend props named:
`isOpen`
`handleOpenChange`
Rather than these longer variants.

src/app/wallet/account-modal.tsx
```diff
@@ -0,0 +1,64 @@
...
+    <Dialog open={accountModalVisible} onOpenChange={handleModalVisible}>
+      <DialogContent
+        className='p-8 sm:rounded-[32px]'
+        hideDefaultCloseButton={true}
+        aria-describedby={undefined}
+      >
+        <DialogTitle>
+          <DialogHeader className='flex w-full flex-row items-start justify-between text-2xl font-semibold'>
+            <p>USDC account</p>
+            <X
```
Cutaiar, line 33: Rather than using icons with custom CSS repeatedly, we can use 
```
<Button variant='ghost' asChild>
    <X ... />
</Button>
```

or something like this to be more consistent. 

I do like the addition of a subtle hover animation here. I would love if we could work this into our shadcn Button rather than custom build it each time like this. Doesn't need to be for this PR.

src/app/wallet/my-accounts.tsx
```diff
@@ -0,0 +1,73 @@
...
+  usdcBalance: string;
+  address: string | null;
+}
+
+export const MyAccounts: React.FC<MyAccountsProps> = ({
+  usdcBalance,
+  address,
+}) => {
+  return (
+    <div className='min-h-[100%] min-w-[360px] rounded-2xl bg-white p-6 shadow-[0_16px_50px_0px_rgba(0,0,0,0.15)]'>
```
Cutaiar, line 17: For this shadow, please use the `shadow-md`, `shadow-lg`, etc. that matches closest to design.

Generally, we want to avoid [tailwind arbitrary values](https://tailwindcss.com/docs/adding-custom-styles#using-arbitrary-values) like hex values and custom shadows like this. Unfortunately, we are forced to make some hard decisions because our in-code color palette does not quite match what our designers are using. Ping me when you have a ~20min tomorrow, I'd love to chat about this so that we can get our mental models aligned.

src/app/wallet/my-accounts.tsx
```diff
@@ -0,0 +1,73 @@
...
+      <div className='flex items-center gap-2'>
+        <span className='rounded-full bg-black p-2 text-white'>
+          <Bank className='size-[1.125rem]' />
+        </span>
+        <span className='text-md font-medium text-[#595B5A]'>MY ACCOUNTS</span>
+      </div>
+      <div className='divider my-4 h-[1px] w-full bg-[#F2F2F2]' />
```
Cutaiar, line 24: what does `divider` do?

src/app/wallet/my-accounts.tsx
```diff
@@ -0,0 +1,73 @@
...
+    setShowAccountModal((prev) => !prev);
+  };
+
+  const handleAccountModalVisible = (open: boolean) => {
+    setShowAccountModal(open);
+  };
+
+  return (
+    <div className='flex gap-2 rounded-xl border border-[#EFEFEF] p-3'>
+      <Image src='/svg/base-usdc.svg' width={35} height={35} alt='usdc logo' />
```
Cutaiar, line 54: recommend adding `priority` prop on this image since it is always in view
https://nextjs.org/docs/pages/api-reference/components/image#priority

Cutaiar (approved):
Nice, speedy work Audrey. I left some comments to address and mentioned pinging me in one of them. Please do message me tomorrow when you have a moment and we can chat. Besides that, this is approved :)

src/app/wallet/my-accounts.tsx
audreyostrom: Ah, this one is on me. I was thought it was a custom CSS class made for borders from looking at `chat-bottombar.tsx`. This might've also been me miswriting `divide`. Either way it'll be removed.

src/app/wallet/my-accounts.tsx
audreyostrom: That makes sense, never knew that was a prop – good to know!

src/app/wallet/my-accounts.tsx
audreyostrom: For sure, whenever you have time. I can even do it right now (didn't see standup was pushed back to 11am). @Cutaiar

src/app/wallet/account-modal.tsx
audreyostrom: Sounds good. I refactored this into its own component as a button for now like you suggested. Since this is a repeated component across multiple modals, I put it the common folder. I can revisit this issue in another PR if there's anything wrong that approach.

src/components/common/x-button.tsx
```diff
@@ -0,0 +1,25 @@
+import { X } from '@untitled-ui/icons-react';
+
+import { Button } from '@/components/ui/button';
+
+interface CloseButtonProps {
+  onClick: () => void;
+  height?: number;
+  width?: number;
```
Cutaiar, line 8:
```suggestion
    className?: string
```

src/components/common/x-button.tsx
```diff
@@ -0,0 +1,25 @@
...
+}
+
+export const CloseButton: React.FC<CloseButtonProps> = ({
+  onClick,
+  height = 6,
+  width = 6,
```
Cutaiar, line 14:
```suggestion
  className
```

src/components/common/x-button.tsx
```diff
@@ -0,0 +1,25 @@
...
+}) => {
+  return (
+    <Button
+      onClick={onClick}
+      variant='ghost'
+      className={`h-${height} w-${width} p-0 transition ease-out hover:scale-110`}
```
Cutaiar, line 20: Rather than extracting css properties into props, a more flexible way to allow callers of this component would be to accept a className prop and merge it with these classNames like `cn(className, 'p-0 ...')

src/components/common/x-button.tsx
```diff
@@ -0,0 +1,25 @@
...
+    >
+      <X className={`h-${height} w-${width} cursor-pointer text-[#98A2B3]`} />
+    </Button>
```
Cutaiar, line 23: Then if this needs to be the same w and h as its parent, you can use `size-full`

src/app/wallet/my-accounts.tsx
```diff
@@ -0,0 +1,83 @@
...
+  address: string | null;
+}
+
+export const MyAccounts: React.FC<MyAccountsProps> = ({
+  usdcBalance,
+  address,
+}) => {
+  return (
+    <div className='min-h-[100%] min-w-[360px] rounded-2xl bg-white p-6 shadow-md'>
+      <div className='flex items-center gap-2'>
```
Cutaiar, line 19: You can achieve this with `min-h-full`

src/app/wallet/account-modal.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+export const AccountModal: React.FC<AccountModalProps> = ({
+  isOpen,
+  handleOpenChange,
+  address,
+}) => {
+  return (
+    <Dialog open={isOpen} onOpenChange={handleOpenChange}>
+      <DialogContent
+        className='p-8 sm:rounded-[32px]'
```
Cutaiar, line 25: how come this is only applied at the `sm` breakpoint? And, dialogs should already be rounded?

src/app/wallet/account-modal.tsx
```diff
@@ -0,0 +1,64 @@
...
+        <div className='mt-6 text-sm text-[#595B5A]'>
+          Ethereum (Base) address
+        </div>
+        <div className='mb-4 flex items-center justify-center gap-2 rounded-xl border border-[#EFEFEF] p-3'>
+          <div className='text-sm text-[#595B5A]'>{address}</div>
+          <div className='ml-auto flex items-center justify-end'>
+            <CopyIconButton
+              copyIconClassName='size-5 text-[#D0D5DD]'
+              checkIconClassName='size-5'
+              onCopy={() => {
```
Cutaiar, line 48: so i provide the onCopy prop in case the caller needs to do something fancy. in this case since you just want to copy the `address`, you can use `stringToCopy={address}`

src/app/wallet/account-modal.tsx
audreyostrom: For this one I was referencing the stylings of previous modals in the codebase, but I can see where this doesn't make sense.

src/app/wallet/account-modal.tsx
```diff
@@ -0,0 +1,64 @@
...
+  DialogTitle,
+} from '@/components/ui/dialog';
+
+interface AccountModalProps {
+  isOpen: boolean;
+  handleOpenChange: (open: boolean) => void;
+  address: string | null;
+}
+
+export const AccountModal: React.FC<AccountModalProps> = ({
```
Cutaiar, line 17: See the discussion here:  https://www.figma.com/design/j8hO0Ewoyi13gZIHcboR1Q?node-id=5741-6575#1036284461

Since the modal only displays the address we are going to nix it for now. Feel free to keep this code on your local (in a stash or branch) though in case we need it in the future.

PR #138, opened 2024-11-20
[SRBT-330] Adjusted wallet container width + added wallet truncation

audreyostrom:
# [SRBT-330] Adjusted wallet container width + added wallet truncation

**Changes**

- Wallet container width has been adjusted to be a maximum of 1080px as per the Figma.
- Wallet Address are now truncated to only show the first three digits and last four digits.

# Screenshots

*Using mock data*
Screenshot 2024-11-20 at 9 09 34 AM

### DIFF
--- a/src/app/wallet/funds-flow.tsx
+++ b/src/app/wallet/funds-flow.tsx
@@ -22,6 +22,7 @@ const BalanceItem: React.FC<BalanceItemProps> = ({
   account,
   balance,
 }) => {
+  const truncatedAccount = `${account.slice(0, 5)}...${account.slice(-5)}`;
   return (
     <div className='flex items-center justify-between'>
       <div className='flex items-center gap-3'>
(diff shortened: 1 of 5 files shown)


src/app/wallet/funds-flow.tsx
```diff
@@ -22,14 +22,15 @@ const BalanceItem: React.FC<BalanceItemProps> = ({
   account,
   balance,
 }) => {
+  const truncatedAccount = `${account.slice(0, 5)}...${account.slice(-5)}`;
```
Cutaiar, line 25: Let's use the same function you define in `transactions-table.tsx`. Perhaps you could move it out to a shared place so that the whole wallet page can use it. This way, there is one source of truth for how wallet addresses are displayed.

Cutaiar (approved):
Nice work!

src/app/wallet/funds-flow.tsx
audreyostrom: sounds good, it's moved to a new file in the wallet folder called `utils.tsx`.

PR #58, opened 2024-08-27
New PR for user not found

ChristopherPark3:
# Description & Technical Solution
- Adds functionality to the `claim my sorbet` button
- Adds a blurred logos to background to use white space
- Also adds `MagicUI` shiny button component

# Video

https://github.com/user-attachments/assets/d79da3a4-0297-4492-8460-1fe786838a58

PR #140, opened 2024-11-21
Invoice Polish

Cutaiar:
**Changes**

- Fix top padding on invoices page
- Allow reopening invoices
- Match invoice table empty state to design
- Autofill invoice numbers
- Prevent users from using the same invoice number twice
- Render overdue badges based on date

Related: https://github.com/MySorbet/sorbet-api/pull/58

# Screenshots


ChristopherPark3 (approved):
Awesome excited to see this in prod!

PR #143, opened 2024-11-28
PostHog analytics

ChristopherPark3:
# PostHog analytics

**Changes**

- Captures widget clicks and sends relevant data to PostHog


Cutaiar:
Making a note here to make sure that clicks from someone in edit mode aren't counted.

PR #144, opened 2024-12-03
Fix: Tooltip Not Appearing for "Add Section Title"

audreyostrom:
# [SRBT-349] Fix: Tooltip Not Appearing for "Add Section Title"

**Changes**

- Refactored the SectionTitleIcon to be wrapped in a `div` instead of an `svg`.  I also refactored the Tooltip Trigger for section titles to be a button. Apparently, Tooltips can only to appear on [buttons](https://github.com/radix-ui/primitives/issues/955#issuecomment-960610209), or elements that require a click / pointer-down interaction – from what I've gathered on the documentation beyond this specific GitHub issue.

# Screenshots

Screenshot 2024-12-03 at 8 57 21 AM

### DIFF
--- a/src/components/profile/widgets/add-widgets.tsx
+++ b/src/components/profile/widgets/add-widgets.tsx
@@ -208,20 +208,14 @@ export const AddWidgets: React.FC<AddWidgetsProps> = ({
             <TooltipProvider>
               <Tooltip>
                 <TooltipTrigger asChild>
-                  <label
-                    className={cn(
-                      'hover:text-sorbet align-center flex cursor-pointer items-center justify-center',
-                      loading && loadingClasses
-                    )}
+                  <Button
+                    variant='ghost'
+                    className='h-fit p-0'
+                    onClick={addSectionTitle}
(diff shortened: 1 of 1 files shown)


src/components/profile/widgets/add-widgets.tsx
```diff
@@ -204,20 +204,30 @@ export const AddWidgets: React.FC<AddWidgetsProps> = ({
...
+                      loading && loadingClasses
+                    )}
+                  >
+                    <input
+                      type='button'
+                      className='hidden'
+                      onClick={addSectionTitle}
+                      disabled={loading}
+                    />
+                    <SectionTitleIcon width={20} height={20} />
```
Cutaiar, line 223: Rather than extracting width and height as props, you can just use `className='size-5'` for a simpler interface

Cutaiar, line 223: See my slack message about using a shadcn button

Cutaiar (approved):
approved some suggestions

PR #145, opened 2024-12-03
[SRBT-343] Wallet Page Skeletons for Loading + Shadow Fixes

audreyostrom:
[SRBT-343] Wallet Page Skeletons for Loading + Shadow Fixes

Also a correction for [SRBT-344](https://linear.app/mysorbet/issue/SRBT-344/consistent-shadows-on-wallet-page-cards) and [SRBT-336](https://linear.app/mysorbet/issue/SRBT-336/wallet-balance-update-delay-may-cause-user-confusion) 

**Changes**

- When the page is loading, skeleton components are rendered instead of the actual cards themselves to reduce confusion about what the actual wallet balance is, etc. In the case of the Wallet Balance Card, there's a skeleton component for just the balance actual value changing (regarding SRBT-336), since the balance loading != page loading necessarily (like in the case of adding funds)
- All cards now have the same box-shadow value (`shadow-md`), instead of arbitrary values

# Screenshots

https://github.com/user-attachments/assets/f3f976ac-47d1-4ab8-b184-29e260a300b7

### DIFF
--- a/src/app/wallet/funds-flow.tsx
+++ b/src/app/wallet/funds-flow.tsx
@@ -3,6 +3,7 @@ import React from 'react';
 
 import { formatCurrency, formatWalletAddress } from '@/app/wallet/utils';
 import { Spinner } from '@/components/common';
+import { Skeleton } from '@/components/ui/skeleton';
 
 interface FundsFlowProps {
   icon: React.ReactNode;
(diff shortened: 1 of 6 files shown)


src/app/wallet/wallet-balance.tsx
```diff
@@ -44,59 +46,66 @@ export const WalletBalance: React.FC<WalletBalanceProps> = ({
...
+                    BALANCE
+                  </span>
+                  {percentChange !== 0 && (
+                    <PercentageChange percentChange={percentChange} />
+                  )}
+                </div>
+                <div className='mt-2 flex'>
+                  {isBalanceLoading || usdcBalance === '' ? (
+                    // Show skeleton with the same height as the balance display
+                    <Skeleton className='h-[36px] w-32 bg-gray-300 leading-[38px]' />
```
Cutaiar, line 71: FYI `h-[36px]` is the same as `h-9`. The latter is preferred because it should be responsive to user set font size. This is the[ best blog post](https://www.joshwcomeau.com/css/surprising-truth-about-pixels-and-accessibility/#accessibility-considerations-5) out there on the topic. 

And then why does the Skeleton need leading? I would not expect it to have an effect on a div (which is what a Skeleton is).

src/app/wallet/wallet-container.tsx
```diff
@@ -4,7 +4,7 @@ import { useFundWallet } from '@privy-io/react-auth';
 import { useSmartWallets } from '@privy-io/react-auth/smart-wallets';
 import { ArrowDown, ArrowUp, Plus } from 'lucide-react';
 import Link from 'next/link';
-import { useCallback, useEffect, useState } from 'react';
+import { useCallback, useEffect, useRef, useState } from 'react';
```
Cutaiar, line 7: Maybe I am missing something but i dont see this used?

src/app/wallet/wallet-container.tsx
```diff
@@ -69,8 +69,9 @@ export const WalletContainer = () => {
           amount: defaultFundAmount,
           asset: 'USDC',
         });
-        setReload(!reload); // trigger reload to refresh wallet amount
       }
+      // TODO: see if setReload() can be removed, fundWallet() is technically finished after just opening the modal.
+      setReload(!reload);
```
Cutaiar, line 74: Yeah I would add a note here that we want to do _something_ to reload, but this is not it.

Cutaiar (approved):
Approved with some suggestions

src/app/wallet/wallet-container.tsx
audreyostrom: Yes, sorry. This was a draft PR, I'll remove the little things like this to be more sensical.

audreyostrom:
Sorry for some of the draft PR flaws! Completely on me from inviting a review before things were fixed.

src/app/wallet/funds-flow.tsx
```diff
@@ -47,55 +48,61 @@ export const FundsFlow: React.FC<
...
-            </span>
-            <span className='text-muted-foreground text-sm font-medium uppercase'>
-              {title}
-            </span>
+    <>
+      {isLoading || balance === '' ? (
+        <Skeleton className='h-44 rounded-3xl bg-gray-300 shadow-md' />
+      ) : (
+        <div className='bg-card flex min-h-full flex-col rounded-3xl p-6 shadow-md'>
+          {/** isLoading && (
```
Cutaiar, line 56: can remove this commented code if it is not used

PR #148, opened 2024-12-16
[SRBT-371] User can only edit section titles when signed in

audreyostrom:
# [SRBT-371] User can only edit section titles when signed in

**Changes**

- I've added additional prop (`editMode`) to section titles widgets, similar to what the other widget types have, to disable a user from editing the text of a section widget if they're aren't authorized to do so.

# Screenshots

https://github.com/user-attachments/assets/aa27647c-29fc-453d-887e-72c632600d41


Cutaiar:
Interesting, there should also be a measure on the backend to only allow an authed user to edit their own profile. This can be merged but lets chat about this too.

Cutaiar (approved):

PR #150, opened 2024-12-16
[SRBT-362] Remove unintended animations after cropping finish

audreyostrom:
# [SRBT-362] Remove unintended animations after cropping finish

**Changes**

- To remove the widgets animating from (0,0) when cropping is finished, I've added `useCSSTransforms={false}` to ReactGridLayout as per this [issue](https://github.com/react-grid-layout/react-grid-layout/issues/103) for the library.

### DIFF
--- a/src/components/profile/widgets/react-grid-layout-custom.module.css
+++ b/src/components/profile/widgets/react-grid-layout-custom.module.css
@@ -7,6 +7,11 @@ Read this to learn more https://stackoverflow.com/a/69541558
 .react-grid-layout-custom {
   /* Isolate the z-index shenanigans below to RGL */
   isolation: isolate;
+  transition: none !important;
+}
+
+.react-grid-item.cssTransforms {
+  transition-property: none !important;
 }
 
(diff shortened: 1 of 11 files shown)


Cutaiar (approved):

Cutaiar (changes requested):
Actually, I'm not sure this is the best way to achieve the desired effect?

```
// Uses CSS3 translate() instead of position top/left.
// This makes about 6x faster paint performance
useCSSTransforms: ?boolean = true,
```

From the docs on RGL, looks like this creates a 6x slowdown? I'm wondering why we get the 0,0 animation after crop but not on initial page load.

audreyostrom:
@Cutaiar I believe the animation might be due to the fact that I hide the react-grid-layout when cropping, and then unhide it. I believe there's some default stylings for that situation. I found a way to address this without using `useCSSTransforms` by finding the styling causing it and disabling it.

Cutaiar:
Awesome, please include some comments in that CSS to call out to future devs why we have that.

Cutaiar (approved):

PR #165, opened 2025-02-21
[SRBT-420] Add POC of First name and Last name in settings

neel-jotaniya:
# [SRBT-420] Add POC of First name and Last name in settings

**Changes**

- Uncomment the endpoint for the settings page.
- Create a basic POC for the First Name and Last Name fields in the settings page.
- Modify the .prettierrc.js file to resolve the end-of-line issue in windows device

# Screenshots
image


src/components/app-sidebar/app-sidebar.tsx
```diff
@@ -1,4 +1,4 @@
-import { LucideIcon } from 'lucide-react';
+import { LucideIcon, SettingsIcon } from 'lucide-react';
```
Cutaiar, line 1: You can import this from `@/components/ui/settings.tsx` for an animated version 👍

src/components/app-sidebar/app-sidebar.tsx
```diff
@@ -134,11 +134,11 @@ const accountItems: MenuItem[] = [
...
-  // {
-  //   title: 'Settings',
-  //   url: '/settings',
-  //   icon: SettingsIcon,
-  // },
+  {
+    title: 'Settings',
+    url: '/settings',
+    icon: SettingsIcon,
+  },
```
Cutaiar, line 141: Lets add a feature flag in `src/lib/flags.ts` called `settings` which is set to `!(process.env.NODE_ENV === 'production)'`. Then use that flag to conditionally add this menu option. Then, we can ship this (by merging it, we will trigger a build and deploy). And you can see it live at `staging.mysorbet.io` and on local. But it will not show up in production (at mysorbet.io)

src/app/settings/page.tsx
```diff
@@ -0,0 +1,93 @@
...
+          <Form {...form}>
+            <form onSubmit={handleSubmit} className='flex flex-col gap-6'>
+              <FormField
+                name='firstName'
+                control={form.control}
+                render={({ field }) => (
+                  <FormItem>
+                    <FormLabel className='h-[14px] block mb-[2px]'>First Name</FormLabel>
+                    <FormControl>
+                      <Input className='h-[36px]' placeholder='Your first name' {...field} />
```
Cutaiar, line 66: In general. try to avoid hardcoding pixel values. instead, use tokens like `h-9` which would still create a 36px height, but would react to the users preferred text size.

In this particular case, we likely don't want to style the input height at all. I would expect it to be consistent with design without extra styling. If this is not the case, let's chat about it and make sure we align with design without having to use extra tokens like this.

src/app/settings/page.tsx
```diff
@@ -0,0 +1,93 @@
...
+  const form = useForm<SettingsFormSchema>({
+    resolver: zodResolver(formSchema),
+    defaultValues: defaultValues || {
+      firstName: '',
+      lastName: '',
+    },
+    mode: 'onTouched',
+  });
+
+  const handleSubmit = form.handleSubmit((data) => {
```
Cutaiar, line 42: The submit here can be very similar to the one found in `src/components/profile/profile-edit-modal.tsx`. And that component will be helpful in general for implementing this page. Feel free to abstract and improve this however you would like (i.e. make a reusable hook, simplify old code, etc.). For this MVP, let's have the fields actually update the user in the DB.

src/app/settings/page.tsx
```diff
@@ -0,0 +1,93 @@
...
+  lastName: z.string().min(1, 'Last name is required'),
+});
+
+type SettingsFormSchema = z.infer<typeof formSchema>;
+
+interface SettingsPageProps {
+  defaultValues?: {
+    firstName: string;
+    lastName: string;
+  };
```
Cutaiar, line 29: Rather than except these as page props, you could use `useAuth` to get the details of the currently logged in user. And use those as default values. Take a look around the repo to see how we do that.

Cutaiar (changes requested):
Great start to this POC, Neel. In general, your code looks great and it seems like this was no problem for you. Take a look at my comments above, mostly minor things that are specific to our codebase. ping me in slack for any questions.

Cutaiar (approved):

src/components/app-sidebar/app-sidebar.tsx
neel-jotaniya: This feature flag is added now

